PR #4, opened 2016-05-05
Subscribe Handles non-Branch links, Universal Object Support, other cleanup

rt2zz:
While these features are not complete nor fully tested, I think it best to get this merged into master on the sooner side so we can get more eyes on this and speed up development


rt2zz:
The key feature in this PR that may take some thought/discussion is the passing through of non-branch deep links to subscribe. The idea is that as a consumer of Branch I do not want to set up parallel deep linking handlers for branch/non-branch links, instead I want to handle all linking through the single subscribe call.

PR #10, opened 2016-05-14
Allow the share sheet to be shown from presented VC's

lprhodes:
Allow the share sheet to be shown from a modally presented view controller. Could be changed to use a while loop if we wish to keep moving up the presentation hierarchy until we have the last one but this is a quick fix for many use cases.


rt2zz:
Can you clarify what issue is fixed by adding a `fromViewController`? And what benefits would there be by moving up the hierarchy?

lprhodes:
At the moment if you were to try and show the share sheet from a modally presented view controller, it would fail. The change fixes this.

rt2zz:
👍

PR #17, opened 2016-06-28
Use base Activity instead of ReactActivity

mehcode:
My setup (at the moment) has me not extending from `ReactActivity` but `Activity` (and implementing all the gunk in there myself; don't ask, it's all in the quest for a fluid splash screen -- react native 0.29 makes what I'm doing unnecessary which is good).

### DIFF
--- a/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
+++ b/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
@@ -1,5 +1,6 @@
 package io.branch.rnbranch;
 
+import android.app.Activity;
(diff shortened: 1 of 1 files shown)


rt2zz:
It is going to take me a minute to ponder this. I suppose there is nothing specific to ReactActivity that we use so this change should have no complications correct?

mehcode:
@rt2zz Things still work great for me after the change. I figure if the project ever needs something in ReactActivity it can change it back. No sense in asking for more than you need now.

rt2zz:
:+1:

PR #21, opened 2016-07-04
fix iOS showShareSheet() linkProperties param being null

athibaud:
just a lil' typo that resulted in `linkProperties` being null.


rt2zz:
🎈 Good catch!

PR #22, opened 2016-07-05
Fixed podspec, added dependency to React, so it can be properly linked

michaelkruglos:
Changes to podspec:

1. Added 'React' as dependency
2. The values in podspec are populated from package.json


rt2zz:
👍  🎈

PR #26, opened 2016-07-10
expose email subject for share sheet

aaustin:
@rt2zz This was [requested on the core library](https://github.com/BranchMetrics/ios-branch-deep-linking/issues/383) and I had forgotten we already do expose it. This makes the interface a bit more consistent

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,7 @@ Control parameters for the link.
 | ------------------ | -------- | --------------------
 | messageHeader      | `string` | The header text
 | messageBody        | `string` | The body text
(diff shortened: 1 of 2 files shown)


rt2zz:
👍 looks great will include in a minor version release early next week.

athibaud:
thanks! though now we have a `messageHeader` prop which is android only that appears as headers on some share patforms (fo eg. Slack) but also as email's subject.. + an iOS only `emailSubject` prop which is only used as email's subject.. maybe we can unify this a little? only expose `messageHeader` and use that as `emailSubject` on iOS for now.. ?

PR #27, opened 2016-07-11
[ios] bring back `channel` and `completed` completion params to showShareSheet()

athibaud:


rt2zz:
Good call, after digging in more it looks like completion block is the recommended format. Opened a corresponding PR here: https://github.com/BranchMetrics/ios-branch-deep-linking/pull/389

PR #29, opened 2016-07-21
[ios] update RCTEventEmitter usage

rt2zz:


rt2zz:
@michaelkruglos I think this undo's your change from https://github.com/BranchMetrics/react-native-branch-deep-linking/pull/25

Upon cursory testing `#import <React/RCTEventEmitter.h>` does not work. Does this syntax require different header search paths?

PR #37, opened 2016-07-27
Update Event Emitter

rt2zz:
**background** the move to stay in line with react-native 0.30 was premature and broke implementation for people stuck on pre 0.28. Instead we will wait until the old event emitter usage is fully deprecated before making the switch.

Leaving this PR open as a reminder and to track progress.

### DIFF
--- a/ios/RNBranch.h
+++ b/ios/RNBranch.h
@@ -1,10 +1,3 @@
-//
-//  RNBranch.h
-//  RNBranch
-//
-//  Created by Kevin Stumpf on 1/28/16.
-//
-
 #import <Foundation/Foundation.h>
 #import "RCTEventEmitter.h"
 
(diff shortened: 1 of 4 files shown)


aaustin:
we don't need to merge this in for people upgrading?

aaustin:
Let me know if you need me to help with anything

rt2zz:
We do not need this yet, until react-native officially deprecates the old method. In this case I think I jumped the gun implementing this (and publishing v0.4) as it is not backwards compatible with react-native <0.28. I published v0.5 today which reverts back to the old method, and opened this PR so when the official deprecation comes we can easily switch over.

rt2zz:
note: there was one backwards compatible change to the header imports required for rn 0.30 which is included in the v0.5 release.

jdee:
I just did this (#152). Closing this PR.

PR #39, opened 2016-08-19
Improve support for non-branch deep linking

rt2zz:
- do not double emit session event
- clear session results upon redemptions, as to avoid duplicate redemption

### DIFF
--- a/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
+++ b/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
@@ -106,8 +106,9 @@ public class RNBranchModule extends ReactContextBaseJavaModule {
   }
 
   @ReactMethod
-  public void getInitSessionResult(Promise promise) {
+  public void redeemInitSessionResult(Promise promise) {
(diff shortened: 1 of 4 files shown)


android/src/main/java/io/branch/rnbranch/RNBranchModule.java
```diff
@@ -106,8 +106,9 @@ public String getName() {
   }
 
   @ReactMethod
-  public void getInitSessionResult(Promise promise) {
+  public void redeemInitSessionResult(Promise promise) {
```
aaustin: will this change the interface for devs and break builds?

rt2zz: `getInitSessionResult` was an "internal" api, i.e. not exposed via the javascript module. 

While it is technically possible to access via `NativeModules.RNBranch.getInitSessionResult` I would say it is unlikely anyone has done this, and additionally it would be considered unsupported module usage: subject to change.

aaustin: Sounds great

aaustin:
👍

PR #42, opened 2016-09-06
fix documentation misprint

eds101:
in both the Java and the Objective C, the key is "metadata" not "contentMetadata".


eds101:
(cost me an extra half hour to get my program working b/c of this misprint -- so please fix in respect for others' time)

rt2zz:
Thank you!

PR #55, opened 2016-11-03
Adding environment switch for iOS (to use the test key instead of the live one)

yazgazan:
The purpose of this PR is to allow users to switch from live to test programatically. For this to work, `branch_key` needs to be a dict with both `live` and `test` keys set.

This would be typically used with a compile-time environment switch of some sort:

```objc
if ([env isEqualToString:@"DEV"]) {
  [RNBranch useTestInstance];
}
[RNBranch initSessionWithLaunchOptions:launchOptions isReferrable:YES];
```


rt2zz:
Thank you this is a useful feature.

I am thinking the documentation sample should use the existing `RCT_DEV` flag to switch between live/test.

Merging now to test drive this, and look into a analogous android side implementation.

PR #56, opened 2016-11-03
Add support for installation via Carthage

gilesvangruisen:
Thanks for this framework!

This PR adds support to be able to install the framework via the [Carthage](https://github.com/Carthage/Carthage) dependency manager.

Essentially, it's just adding the `Carthage` build directory to the target's Framework Search Paths, and adding the framework's `Headers` directory to the target's Header Search Paths. Neither of these changes should have any affect on existing and future CocoaPods integrations, just as including the `Pods` directories in Framework/Header search paths has no affect on Carthage integrations.

I have tested this fork on a private project, and it installs and builds just fine, but feel free to do your own testing.


rt2zz:
great work, thank you very much!

I am going to amend the readme a bit, but 👍

PR #61, opened 2016-12-07
Fixes "'Branch/Branch.h' not found" errors.

theseansy:
This fixes issues when Branch is installed via cocoapods, when React Native is *not* installed via cocoapods.

See issues:
https://github.com/BranchMetrics/react-native-branch-deep-linking/issues/60,  https://github.com/BranchMetrics/react-native-branch-deep-linking/issues/18 for more information.  The final solution by @Morhaus is hard to maintain, as RNBranch.m needs to be adjusted after each fresh `npm install`.

Because `Branch.h` in `Pods` is actually under `Branch-SDK`, `Branch/Branch.h`. Thus just importing `"Branch.h"` works well with my configuration, given the `HEADER_SEARCH_PATHS` in this PR. However, I haven't verified in the case where both React and branch are installed via cocoapods.

It seems like the `HEADER_SEARCH_PATHS` entry `$(SRCROOT)/../../../ios/Pods/Headers/Public/Branch` **should have** solved this problem, but after `pod install` with the latest versions of `cocoapods` and `Branch`, there aren't any Header files that I find in that `/Headers` directory, so it doesn't seem to help.

### DIFF
--- a/ios/RNBranch.m
+++ b/ios/RNBranch.m
@@ -2,7 +2,7 @@
 #import "RCTBridgeModule.h"
 #import "RCTBridge.h"
 #import "RCTEventDispatcher.h"
-#import <Branch/Branch.h>
+#import "Branch.h"
 #import "BranchLinkProperties.h"
 #import "BranchUniversalObject.h"
 
--- a/ios/RNBranch.xcodeproj/project.pbxproj
+++ b/ios/RNBranch.xcodeproj/project.pbxproj
@@ -230,6 +230,7 @@
 					"\"$(SRCROOT)/../../../ios/Pods/Headers/Public\"",
 					"\"$(SRCROOT)/../../../ios/Pods/Headers/Public/Branch\"",
 					"\"$(SRCROOT)/../../../ios/Carthage/Build/iOS/Branch.framework/Headers\"",
+					"\"$(SRCROOT)/../../../ios/Pods/Branch/**",
 				);
 				LIBRARY_SEARCH_PATHS = (
 					"$(inherited)",
@@ -252,6 +253,7 @@
 					"\"$(SRCROOT)/../../../ios/Pods/Headers/Public\"",
 					"\"$(SRCROOT)/../../../ios/Pods/Headers/Public/Branch\"",
 					"\"$(SRCROOT)/../../../ios/Carthage/Build/iOS/Branch.framework/Headers\"",
+					"\"$(SRCROOT)/../../../ios/Pods/Branch/**",
 				);
 				LIBRARY_SEARCH_PATHS = (
 					"$(inherited)",


rt2zz:
Thanks for putting this together. This is going to take some comprehensive testing. the `<Branch/Branch.h>` import was implemented to solve a swift import issue.

I have an app with the same setup as you, Branch as pod, rn and rn-branch both as npm modules, and headers resolve fine. I will have to try isolating this issue in the testbed.

theseansy:
@rt2zz - thanks for the info.  It's good to know you have the same setup and it works for you.

I have a swift 2.3 code in my project, and this still compiles with this change.  I did see someone said if you remove `use_frameworks!` from the Podfile, that got it working for him/her, but this makes swift pods incompatible for `pod install`, so I can't do that.

In my project `HEADER_SEARCH_PATHS` I have `$(SRCROOT)/../node_modules/react-native-branch/ios`, but that allows the project to find `RNBranch.h`.  Without changing the HEADER_SEARCH_PATH from RNBranch.xcodeproj, I don't see how this will can resolve the path to the Branch framework, from a Pod or manually included Branch framework.

`#import <Branch-SDK/Branch.h>` also works by the way, but this is because the Pod folder name is `Branch-SDK`.

good2hear:
@theseansy thanks for your work!! I was having the same issue..
just took a look at the changes you made.
but in my case, when I added 

`$(SRCROOT)/../../../ios/Pods/Branch/**` 
in `HEADER_SEARCH_PATH` from `RNBranch.xcodeproj` , 

I was still getting Branch.h file not found.
but when I got rid of the ** at the end (like below)

`$(SRCROOT)/../../../ios/Pods/Branch` 

It worked fine.. any idea why this is happening?

theseansy:
@good2hear - When you say you set it to `$(SRCROOT)/../../../ios/Pods/Branch`, did you have it set to `(recursive)` or `(non-recursive)` in the XCode editor?  If it is set to `recursive`, this is how mine is configured.  `recursive` in the XCode **Build Settings** editor evaluates to `/**` in the text path when compilation is run, but XCode tries to abstract this from the user.  If you put in `/**` explicitly in a path in the **Build Settings** editor in XCode, it won't interpret this correctly.  Instead, you must use their `recursive` vs `non-recursive` setting.

good2hear:
@theseansy thanks a lot for reply!! I set it up with `(recursive)`  and it worked! thought you added `/**` explicitly..

any update when this might get merged?

theseansy:
@good2hear - The maintainers need to test this more, but I guess upvote this PR with a 👍  so we can start getting a signal of who is experiencing this issue.

good2hear:
@theseansy just did!! thanks!! : ))

jdee:
Hi @theseansy, @good2hear 

I think this PR will be superseded by changes to be merged from another branch. See the [updates](https://github.com/BranchMetrics/react-native-branch-deep-linking/tree/updates) branch in this repo. That's not yet released nor entirely finished. However, I think those are the necessary changes. It's not ideal to use search paths when it's possible to avoid them. What was really missing was a Copy Headers build phase in the RNBranch.xcodeproj.

The latest changes require react-native 0.40.0, and all headers need to be imported with framework-style imports:

```Objective-C
#import <React/RCTRootView.h>
#import <RNBranch/RNBranch.h>
#import <Branch/Branch.h> # if you're using the Branch SDK directly in Objective-C
```

These changes should be merged soon. But you may be able to build from that branch and verify that it solves your problem.

jdee:
Also please note that `RNBranch.xcodeproj` needs to be after all the React projects when you include it in your own project.

theseansy:
Thanks @jdee.

What do you mean here?
> Also please note that RNBranch.xcodeproj needs to be after all the React projects when you include it in your own project.

Do you mean it's at the very bottom of the project list of `RCT*.xcodeproj` in the Xcode GUI?

jdee:
@theseansy Yes. Like this:

screen shot 2017-01-09 at 4 51 26 pm

Also, please note I updated the previous comment. Please use

```Objective-C
#import <react-native-branch/RNBranch.h>
```

This is necessary for consistency with other integration methods. Sorry for the bad info earlier.

I'm working on getting all installation methods (manual, Cocoapods, Carthage) represented in the sample code and thoroughly documented. For now, this should help with the problem you reported.

This PR https://github.com/BranchMetrics/react-native-branch-deep-linking/pull/67 should be merged soon.

theseansy:
@jdee - Testing your changes, I still get the same issue.  Anything including `<Branch/*>` doesn't resolve within the `RNBranch` files

jdee:
@theseansy I'm glad you can get around it, but that shouldn't be necessary and probably isn't the best approach for the RNBranch project. In case you're interested, there's an extensive explanation below. I may just have to fix something else, or maybe there's something unusual in your setup that I need to account for. Or maybe you have an outdated version of some component. Also, I haven't yet covered all possible methods of installing the Branch SDK. It sounds like you're using Cocoapods for that though, in the same way as the testbed project.

Here are some things to check. 

1. Does RNBranch.xcodeproj already have a (non-recursive) header search path $(PODS_ROOT)/Headers/Public, like this?

screen shot 2017-01-10 at 11 27 18 am

If not, then maybe you just need to make sure you're getting the right RNBranch.xcodeproj from node_modules/react-native-branch/ios.

I've had some issues with yarn caching. If you're using yarn, it could be that a `yarn cache clean` and then running yarn again, possibly first removing node_modules, might fix it.

2. Does your PODS_ROOT contain Headers/Public/Branch?

```bash
[jdee@Jimmys-MacBook-Pro Public (master)]$ pwd
/Users/jdee/github/BranchMetrics/react-native-branch-deep-linking/testbed/ios/Pods/Headers/Public
[jdee@Jimmys-MacBook-Pro Public (master)]$ ls -la
total 0
drwxr-xr-x   3 jdee  staff   102 Jan 10 11:26 .
drwxr-xr-x   4 jdee  staff   136 Jan 10 11:26 ..
drwxr-xr-x  46 jdee  staff  1564 Jan 10 11:26 Branch
```

If not, then maybe you just need to update cocoapods or do a fresh `pod install` or `pod update`. Or maybe there's something in your Podfile or other configuration that is putting something in a different place. If this turns out to be the problem, would you mind sharing your Podfile and any arguments passed to `pod install`?

3. Is your Pods subdirectory a direct child of the ios subdirectory, i.e., YourApp/ios/Pods, next to Yourapp.xcodeproj?

If not, then I may have to do something fancier to detect where it is or get Cocoapods to set up the RNBRanch.xcodeproj when you run `pod install` (or you could consider moving it). But if $(PODS_ROOT)/Branch is picking up the right headers, then this is unlikely to be the problem. The RNBranch.xcodeproj assumes PODS_ROOT is `$(SRC_ROOT)/../../../ios/Pods`. SRC_ROOT for RNBranch.xcodeproj is `node_modules/react-native-branch/ios`.

You can also see the testbed project in the repo to see how it's set up.

#### Here's why

For the last decade or more, in C and C++, it's been standard practice not to include header search paths via `-I` arguments to the compiler for external dependencies. Instead, headers are all installed under subdirectories of a common location. RPM installs headers to /usr/include. Many installers install headers to /usr/local/include. And so on. You always do something like `#include <package/header.h>`, and the package headers are installed in /usr/include/package, for example, and at most the compiler gets a small number of search paths like `-I /usr/local/include` rather than one for each external package.

Objective-C has been slow to catch up to this practice, but this is the current standard practice with frameworks, pods, etc. If you have to add a special search path for a specific dependency, then that dependency is probably broken, and in this case maybe it still is. And there should ordinarily be no need to specify a recursive search path. The header is supposed to be in one place and probably shouldn't be picked up if it's not in the right place.

There are a number of benefits to doing things this way, including compilation time, by keeping search paths shorter, and predictable dependency organization. Another benefit is independence of the dependency installation mechanism. This is the main reason this project had issues before. It has to support a number of possible installation options for the Branch SDK and the native React framework used by Xcode, and the same #imports have to work in all cases.

You can see this as well by running `react-native init TestProject` using react-native 0.40.0. Look at the generated project and the projects included in the Libraries group. None of them has custom search paths to pick up the others. They build cleanly without assistance.

Ordinarily headers are installed to intermediate locations by Xcode or Cocoapods. Cocoapods just symlinks all headers into $(PODS_ROOT)/Headers.

For the React Libraries included by `react-native init`, a Copy Headers phase in each of the Xcode projects installs the headers under DerivedData, e.g.:

```bash
[jdee@Jimmys-MacBook-Pro Public (master)]$ ls -la ~/Library/Developer/Xcode/DerivedData/testbed-agpragiweulurccqrhhswabokldo/Build/Products/Debug-iphonesimulator/include
total 0
drwxr-xr-x    8 jdee  staff   272 Jan 10 11:42 .
drwxr-xr-x   23 jdee  staff   782 Jan 10 11:43 ..
drwxr-xr-x    3 jdee  staff   102 Jan 10 11:42 RCTImage
drwxr-xr-x  124 jdee  staff  4216 Jan 10 11:42 React
drwxr-xr-x    3 jdee  staff   102 Jan 10 11:42 cxxreact
drwxr-xr-x    4 jdee  staff   136 Jan 10 11:42 jschelpers
drwxr-xr-x    3 jdee  staff   102 Jan 10 11:42 react-native-branch
drwxr-xr-x    5 jdee  staff   170 Jan 10 11:42 yoga
```

This is where the #imports resolve for those dependencies.

I think Facebook does this with the projects generated by `react-native init` in order not to force a JS developer to do a `pod install` or something else just to get the native React libraries. Unfortunately, Xcode doesn't run the Copy Headers phase for all dependencies at once. Hence order is important, and RNBranch.xcodeproj has to come after React.xcodeproj in order to pick up the <React/*> headers.

theseansy:
@jdee thanks for the super thorough response.  I look into this and let you know what I changed to get things working (hopefully).  Really appreciate this.

theseansy:
@jdee - so the problem is my Podfile has `!use_frameworks` in it because I use Swift Pods.

If I take out the `!use_frameworks` declaration, I get the following error from cocoapods:

```
[!] Pods written in Swift can only be integrated as frameworks; add `use_frameworks!` to your Podfile or target to opt into using it.
```
Thus unless I only use objective C pods, I have to keep it in there.

jdee:
@theseansy Thanks for that. I'm working on it. I'll have something for you soon. Thanks for your patience.

jdee:
@theseansy Please try the master branch again when you have time. I've modified RNBranch.xcodeproj to pick up the Branch.framework from Cocoapods. There are now several testbed implementations under the testbed directory. The (perhaps poorly-named) testbed_manual project, which is closest to your situation, now has `use_frameworks!` in its Podfile. You can also comment out that line, do a fresh `pod install`, clean the workspace and run it with static libraries instead. It works either way now.

theseansy:
@jdee - I tested it and it worked!  Thanks so much for all of the help.

jdee:
@theseansy Thanks for confirming. I hope we'll be doing a new release by the end of the week.

Issue #84, opened 2017-01-23
Alias not working


timursadykhov:
Hello,

I'm trying to customize the actual link that is created by branch when I show the share sheet. Currently it's just a short url like `http://go.example.com/Abcd/[branch_uuid]` but what i'd like it to be is `http://go.example.com/event/[my_id]`. 

I tried to accomplish this by making linkProperties: 
`let linkProperties = { alias: 'http://go.example.com/event/test' };` 
and 
`let linkProperties = { alias: 'go.example.com/event/test' };` 
but neither seemed to do anything. 

How would I go about doing this?

I'm on **version 0.9.0** and I haven't seen anything that would suggest the feature was not available in this version.

Thank you,
Timur

jdee:
Hi @timursadykhov. It looks like this was an unintentional omission. For example, the word "alias" does not occur anywhere in the native RNBranch code for either platform. Also missing: campaign, stage, tags. I've already fixed this on the iOS side. I have a little more work to do on Android. We'll be doing a release 0.9.1 this week. There is currently a `release-0.9.1` branch in this repo. I should have the fix for your issue in that branch by the end of the day for you to try. I'll notify you when it's ready.

timursadykhov:
Ah I see. Thanks for the quick response, looking forward to the fix!

jdee:
@timursadykhov Those changes have been merged to `release-0.9.1`. I can generate a short URL with an alias. Please try it out when you have a chance. Thanks.

```js
import branch from 'react-native-branch'

buo = branch.createBranchUniversalObject("abc", { title: "A thing", contentDescription: "It is a thing."})
let linkProperties = { alias: "thing2", tags: [ "a_tag", "another_tag"] }
let controlParams = { "$desktop_url": "https://example.com", "$match_duration": 3600 }

buo.generateShortUrl(linkProperties, controlParams).then((response) => {
  console.log("Short URL: " + response.url)
})
```

timursadykhov:
Thanks for the quick turnaround @jdee!

Initially I was unable to build because line 110 in RNBranch.m `linkProperties.campaign = linkPropertiesMap[@"campaign"];` was throwing the error **Property 'campaign' not found on object of type 'BranchLinkProperties** 

I removed that line and was able to build fine. 

The changes work when passing alias to `generateShortUrl` however using it with `showShareSheet` results in a long link with a bunch of params. 

[...]

PR #87, opened 2017-01-26
Portability 0.9 (iOS)

jdee:
This change is intended to help out in cases like #84. RN users are not necessarily familiar with CocoaPods or Carthage. It's easy to run `pod install`, e.g., get an old version of the SDK and have build problems for reasons that are not obvious (though in that case we can help out in the docs by recommending `pod install --repo-update`). It's also easy to forget to update the native SDK when updating the react-native-branch module. In the long run, I'd like to bundle the Branch SDK in with the native RNBranch code in order to avoid a) the necessity to even install CocoaPods or Carthage in order to use react-native-branch; b) the need to document different installation options; c) any possibility of a version mismatch. This closely resembles the way that Facebook bundles all native code with the NPM module to the point of discontinuing the React pod in CocoaPods, so that the only way to get to it is to use the version in the NPM module, which is always the right version. There's otherwise no way to require a specific native SDK version with a specific NPM package version. However, that is likely to be a breaking change, have other consequences and require a lot of testing.

The changes in this branch for 0.9.1 (and later 1.0.1) allow any known property of BranchLinkProperties or BranchUniversalObject to be passed in safely. A warning will be generated in the NSLog at runtime. The user who reported #84 had version 0.12.4 of the Branch iOS SDK installed. With these changes, they would have seen:

```
"alias" is not supported by the installed version of the native Branch SDK for objects of type BranchLinkProperties. Please update to the current release using "pod update" or "carthage update".
```

This ought to help others avoid the same kinds of issues.

The same approach is used for link properties and universal object properties. Note that many properties not supported by 0.9 or 1.0 are currently listed in the README for this repo. This change supports all documented fields in 0.9.1.

### DIFF
--- a/ios/BranchLinkProperties+RNBranch.h
+++ b/ios/BranchLinkProperties+RNBranch.h
@@ -14,5 +14,4 @@
 
 - (instancetype)initWithMap:(NSDictionary *)map;
 
-+ (NSDictionary<NSString *, RNBranchProperty *> *) supportedProperties;
 @end
--- a/ios/BranchLinkProperties+RNBranch.h
+++ b/ios/BranchLinkProperties+RNBranch.h
@@ -3,7 +3,7 @@
 //  RNBranch
 //
 //  Created by Jimmy Dee on 1/26/17.
-//  Copyright © 2017 Dispatcher. All rights reserved.
+//  Copyright © 2017 Branch Metrics. All rights reserved.
 //
 
 #import <Branch/Branch.h>
--- a/ios/BranchLinkProperties+RNBranch.h
+++ b/ios/BranchLinkProperties+RNBranch.h
@@ -8,8 +8,11 @@
 
 #import <Branch/Branch.h>
 
+@class RNBranchProperty;
+
 @interface BranchLinkProperties(RNBranch)
 
 - (instancetype)initWithMap:(NSDictionary *)map;
 
++ (NSDictionary<NSString *, RNBranchProperty *> *) supportedProperties;
 @end
--- a/ios/BranchLinkProperties+RNBranch.h
+++ b/ios/BranchLinkProperties+RNBranch.h
@@ -1,0 +1,15 @@
+//
+//  BranchLinkProperties+RNBranch.h
+//  RNBranch
+//
+//  Created by Jimmy Dee on 1/26/17.
+//  Copyright © 2017 Dispatcher. All rights reserved.
+//
+
+#import <Branch/Branch.h>
+
+@interface BranchLinkProperties(RNBranch)
+
+- (instancetype)initWithMap:(NSDictionary *)map;
+
+@end
--- a/ios/BranchLinkProperties+RNBranch.m
+++ b/ios/BranchLinkProperties+RNBranch.m
@@ -13,7 +13,7 @@
 
 - (instancetype)initWithMap:(NSDictionary *)map
 {
-    self = [super init];
+    self = [self init];
     if (self) {
         [RNBranchProperty setLinkPropertiesOn:self fromMap:map];
     }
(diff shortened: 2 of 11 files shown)


aaustin:
@ahmednawar Can you review Jimmy's PR here? Really appreciate it!

jdee:
I think the refactoring and other general improvements in this branch are a good idea. I'm not so sure about the dynamic aspect. It may be counterproductive to try and support older SDK versions, especially since something similar would have to be done for Android. It would not be hard to revert to a simpler direct mapping and simply document the need to keep the Branch SDK up to date.

ios/BranchLinkProperties+RNBranch.m
```diff
@@ -0,0 +1,44 @@
...
+{
+    static NSDictionary<NSString *, RNBranchProperty *> *_linkProperties;
+    static dispatch_once_t once = 0;
+    dispatch_once(&once, ^{
+        _linkProperties =
+        @{
+          @"alias": [RNBranchProperty propertyWithSetterSelector:@selector(setAlias:) type:NSString.class],
+          @"campaign": [RNBranchProperty propertyWithSetterSelector:@selector(setCampaign:) type:NSString.class],
+          @"channel": [RNBranchProperty propertyWithSetterSelector:@selector(setChannel:) type:NSString.class],
+          // @"duration": [RNBranchProperty propertyWithSetterSelector:@selector(setMatchDuration:) type:NSNumber.class], // deprecated
```
ahmednawar: curious, why is this deprecated?

ios/BranchUniversalObject+RNBranch.m
```diff
@@ -0,0 +1,69 @@
...
+    static dispatch_once_t once = 0;
+    dispatch_once(&once, ^{
+        _universalObjectProperties =
+        @{
+          @"canonicalUrl": [RNBranchProperty propertyWithSetterSelector:@selector(setCanonicalUrl:) type:NSString.class],
+          @"contentDescription": [RNBranchProperty propertyWithSetterSelector:@selector(setContentDescription:) type:NSString.class],
+          @"contentImageUrl": [RNBranchProperty propertyWithSetterSelector:@selector(setImageUrl:) type:NSString.class],
+          @"contentIndexingMode": [RNBranchProperty propertyWithSetterSelector:@selector(setContentIndexingMode:) type:NSString.class],
+          @"metadata": [RNBranchProperty propertyWithSetterSelector:@selector(setMetadata:) type:NSDictionary.class],
+          @"title": [RNBranchProperty propertyWithSetterSelector:@selector(setTitle:) type:NSString.class]
```
ahmednawar: Currently, there are more properties for Branch Universal Object in the native SDK (https://github.com/BranchMetrics/ios-branch-deep-linking/blob/master/Branch-SDK/Branch-SDK/BranchUniversalObject.h) Are there plans to add them later?

ios/BranchLinkProperties+RNBranch.h
```diff
@@ -0,0 +1,17 @@
+//
+//  BranchLinkProperties+RNBranch.h
+//  RNBranch
+//
+//  Created by Jimmy Dee on 1/26/17.
+//  Copyright © 2017 Dispatcher. All rights reserved.
```
ahmednawar: Can you remove the Copyright statement or change it to Branch?

ios/BranchUniversalObject+RNBranch.m
jdee: I'd like to support all parameters. I was starting with the ones listed in the README repo, which were not previously supported.

ios/BranchLinkProperties+RNBranch.h
jdee: I'll change it. I didn't notice that.

ios/BranchLinkProperties+RNBranch.m
jdee: I don't know why duration is deprecated (or if it really is), but it says so [here](https://dev.branch.io/getting-started/configuring-links/guide/#other-parameters). It's easy to support it if that's wrong.

PR #92, opened 2017-02-01
Update documentation to include ~id param

parshap:
I noticed an `~id` param is included but is not mentioned in the readme. I'm actually not sure if *branch link id* is the right way to describe this id. Is that correct? Is this the same thing as the "branch link id" and "branch match id"?

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,7 @@ Branch returns explicit parameters every time. Here is a list, and a description
 
 | **Parameter** | **Meaning**
 | --- | ---
+| ~id | The branch link id
 | ~channel | The channel on which the link was shared, specified at link creation time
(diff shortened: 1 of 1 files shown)


jdee:
@parshap Thanks for the PR. The full list of parameters you can receive is [here](https://dev.branch.io/getting-started/deep-link-routing/guide/react/#branch-provided-data-parameters-in-callback). I'm just at the point of redoing a lot of the docs in this repo, so everything will moving around and changing anyway. As far as I'm aware, everything here is currently correct, but some things like this may be incomplete for the moment. If you don't mind, I'll close this and add the wording from the docs site and a link to the full list at that location.

parshap:
Sounds good! I wasn't aware of the docs site until someone else pointed me to it. A link would definitely be helpful.

PR #93, opened 2017-02-02
Ensure all headers are project to allow iOS archiving

mattoakesnet:
The library will currently prevent you submitting your app to the App Store because it includes a public header. Because of this, Xcode will think the generated archive is "generic" and will not allow you to upload it to iTunes Connect.

This PR fixes this by making all headers "project headers".

See here for more information:

* http://stackoverflow.com/a/13620628/99582
* https://developer.apple.com/library/content/technotes/tn2215/_index.html#//apple_ref/doc/uid/DTS40011221-CH1-HD1


jdee:
@mattoakesnet Thanks for the PR. Apologies for the problem. We'll probably be releasing 0.9.2/1.0.2 in the next couple of days. This will be included.

PR #106, opened 2017-02-16
Add manually imported Branch.framework to the Framework Search Paths

kfiroo:
We don't use Cocoapods or Carthage in our project, so we manually imported the `Branch.framework` and had to fix the `Framework Search Paths` in `RNBranch.xcodeproj`.

Every time we update the package version the it resets this setting.

Can we have it in the default configuration?

Thanks :)

### DIFF
--- a/ios/RNBranch.xcodeproj/project.pbxproj
+++ b/ios/RNBranch.xcodeproj/project.pbxproj
@@ -286,6 +286,11 @@
 		B3A3CC401C5B0A070016AC52 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				FRAMEWORK_SEARCH_PATHS = (
+					"$(CARTHAGE_ROOT)/Build/iOS",
+					"$(PODS_CONFIGURATION_BUILD_DIR)/Branch",
+					"$(SRCROOT)/../../../ios",
+				);
 				OTHER_LDFLAGS = "-ObjC";
 				PRODUCT_NAME = "react-native-branch";
 				PUBLIC_HEADERS_FOLDER_PATH = "/usr/local/include/$(PRODUCT_NAME)";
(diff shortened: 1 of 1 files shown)


jdee:
@kfiroo Thanks for the PR. This is something I've wanted to take care of, but I have some concerns:

It looks like you chose to put the Branch SDK directly next to your main xcodeproj file. That's perfectly fine, but when you manually install the SDK, you're kind of freestyling. This setting forces other users to put it in exactly the same place. Maybe we can make that a recursive search path, so that it will find the Branch SDK anywhere under the ios directory.

I think I'd prefer to add this setting at the project level rather than the target level. The Carthage and CocoaPods settings are already in the project settings. I think it makes more sense just to add the $(SRC_ROOT) setting there instead.

I'm also working on another solution that will eliminate the need to install the SDK separately, so I think this is only a temporary fix anyway. I should have a branch to push soon that you can test out. I suppose this is only a problem for you when we do releases, so as long as we have it in the next release, you shouldn't have any more problems.

You're welcome to make those changes, or I can just do those things myself and close this PR. Either way, we're accepting your changes at least in spirit. Branch would like to offer you a gift for the contribution. Please contact integrations@branch.io to receive your just reward.

kfiroo:
@jdee Hey, I see what you're saying about _freestyling_ it, I'm new to all the xcode settings scene :)

I think, if you're working on a version that will import the framework automatically it's a far better solution! (I was trying to save some headache to others like me who install the framework manually)

I can keep the changes locally until the next (or next next) version is out with better support for this.
Others can still use this PR as a source of information at least :)

jdee:
Hi @kfiroo. See https://github.com/BranchMetrics/react-native-branch-deep-linking/pull/109, which has been merged to the master branch. I think that should fix your issue and work for others as long as they put the framework anywhere under the ios subdirectory, which seems reasonable. That will be part of the next release, which will also include full support for non-Branch links. That should be out this week.

In 1.1, we should be able to include the native Branch SDK. I hope to get that out in the next few weeks.

If your issue is addressed in the master branch, I think we ought to close this PR. It will still show up in searches for this problem, so it will still be helpful to others. And you should contact Branch for your swag. Thanks for making this happen.

kfiroo:
@jdee Amazing! thanks :)

PR #120, opened 2017-03-09
Age cached BUOs

jdee:
This is a remedy to the growth issue discussed in a [recent pull request](https://github.com/BranchMetrics/react-native-branch-deep-linking/pull/117).

Introduced io.branch.rnbranch.AgingHash (Android) and RNBranchAgingDictionary (iOS). These containers require a TTL at initialization and will purge old entries on insertion. Items that have not been accessed within the TTL are removed from the container whenever a new item is inserted. This prevents long-term, unlimited memory growth.

The original problem is that the native layers never know if the corresponding JS object still exists, so it's possible to make a call after an expired Branch Universal Object has been removed from the native cache, if the JS object is still around. Then the specified ident is not found in the cache, and an error has to be generated. In this case an error is returned to JS with the code `RNBranch::Error::BUONotFound`. The JS object still has the parameters used to create the BUO. When it receives this error code, it creates a new BUO on the spot by passing the parameters back to `RNBranch.createUniversalObject`, receives a new ident, updates the ident on the JS object and then calls the method on the new native BUO. The new BUO is inserted into the native cache with a new access time and will be retained for the TTL or until `release()` is called from JS.

This solution has the following virtues:

- The JS BUO wraps a native instance, guaranteeing a one-to-one relationship as long as the BUO is cached.
- The potential memory growth is limited and probably not a significant issue in a well-behaved app to begin with.
- The optional `release()` method is provided for the JS app to do proper clean-up if memory growth is a concern.
- If `release()` is not called on the JS BUO, the native object will not be cached beyond the TTL if further BUOs are created.
- If a cached BUO expires, JS operations will complete normally, at the expense of continuity between native instances.

It also has a couple of potential points of failure. If a call is made after a BUO expires from the native cache, a new instance of the object has to be created, which is the problem that the original change was intended to fix. However, this should happen much less often now. For example, the TTL in the current code base is one hour. To provoke this condition, the app would have to do the following:

- Create a BUO in JS
- Live for more than an hour, probably in the background
- Create a new BUO in JS after the TTL expires for the first BUO
- Use the first BUO again (call a method on the JS BUO)

The most likely scenario here would be something like opening a modal an hour later that creates a new BUO, then dismissing it and calling something like `registerView()` again. In this case, creating a new BUO instance for the same item should not be an issue.

This scenario can also arise if an app incorrectly calls a method on a JS BUO after calling `release()`. Now that automated GC of a sort is available, the `release()` method could be removed. 

There is little defense against certain extreme pathological cases, such as an app creating thousands of temporary BUOs in a loop without calling `release()` each time. In a normal, well-behaved app, there should never be more than a handful of cached BUOs at any given time, and at any rate, the SDK should never fail given that the native BUO can be reconstructed as necessary.

### DIFF
--- a/android/src/main/java/io/branch/rnbranch/AgingHash.java
+++ b/android/src/main/java/io/branch/rnbranch/AgingHash.java
@@ -7,7 +7,7 @@ import java.util.Iterator;
  * Created by jdee on 3/8/17.
  */
 
-public class AgingHash<KeyType, ValueType> extends Object {
+public class AgingHash<KeyType, ValueType> {
     private long mTtl;
     private HashMap<KeyType, AgingItem<ValueType>> mHash = new HashMap<>();
 
--- a/android/src/main/java/io/branch/rnbranch/AgingHash.java
+++ b/android/src/main/java/io/branch/rnbranch/AgingHash.java
@@ -8,11 +8,15 @@ import java.util.Iterator;
  */
 
 public class AgingHash<KeyType, ValueType> {
-    private long mTtl;
+    private long mTtlMillis;
     private HashMap<KeyType, AgingItem<ValueType>> mHash = new HashMap<>();
 
-    public AgingHash(long ttl) {
-        mTtl = ttl;
+    public AgingHash(long ttlMillis) {
+        mTtlMillis = ttlMillis;
+    }
+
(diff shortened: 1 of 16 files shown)


ahmednawar:
Looks great! 👍

PR #125, opened 2017-03-15
Gradle updates and misc.

jdee:
Updated the `com.facebook.react:react-native' requirement in build.gradle to use whatever is in node_modules, rather than pinning to an old version, following the example in the app's build.gradle (generated by RN). AndroidStudio also updated the buildToolsVersion.

Changed the error message generated by the native layers when a BUO is not found, since this always caught in the JS and retried. Also removed some outdated comments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,12 +55,14 @@ Please see the Branch [SDK Integration Guide](https://dev.branch.io/getting-star
 
 ## Usage
 ```js
-import branch, { AddToWishlistEvent,
+import branch, {
+  AddToWishlistEvent,
   PurchasedEvent,
   PurchaseInitiatedEvent,
   RegisterViewEvent,
   ShareCompletedEvent,
-  ShareInitiatedEvent } from 'react-native-branch'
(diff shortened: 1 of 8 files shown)


farazs:
curious why you updated the build tools version? this just breaks it for people who don't have build-tools 25 installed on their machine. 23.0.1 is required by react-native so everyone would have it installed.

we use continuous integration and delivery and this broke our build requiring us to add installing build tools 25.0.0 to our workflow. seems like an unnecessary change.

jdee:
@farazs Sorry for the trouble. That setting was automatically updated by Android Studio, which also harasses you continually until you update. If this causes trouble, it's not hard to revert and push out changes.

farazs:
I've updated our build so it's not a problem for me anymore but others might run into the same issue.

jdee:
Thanks for the feedback. If there are more complaints, I can shoot out a patch. Usually not having the latest build tools breaks my builds. I'll investigate and see what's wrong there.

PR #147, opened 2017-04-05
[OS-35] Android intent

jdee:
This parallels a change in the iOS layer to support link routing in native apps that include RN components. Because the initSession method in the RNBranchModule swallows the callback to Branch.initSession, there has been no direct way for a native Android app to receive callbacks when a link is opened once it integrates react-native-branch. The RNBranchModule was already generating a local broadcast intent in this case, which it alone caught to prompt it to retransmit a stored response (initSessionResult) to the RN app. This event, in turn, could only be caught by JS.

To allow the native layer to be notified also, the local broadcast intent (RNBRanchModule.NATIVE_INIT_SESSION_FINISHED_EVENT) was made public. The intent also now includes a number of extras:

NATIVE_INIT_SESSION_FINISHED_EVENT_ERROR: String - This is getMessage() from the BranchError, if any.
NATIVE_INIT_SESSION_FINISHED_EVENT_URI: Uri - This is the original URI opened by the app, as in the JS.
NATIVE_INIT_SESSION_FINISHED_EVENT_PARAMS: String - This is the original referring params, serialized as JSON.
NATIVE_INIT_SESSION_FINISHED_EVENT_BRANCH_UNIVERSAL_OBJECT: BranchUniversalObject - If `params != null && params.has("~id")` this is ~~a BUO constructed from the params~~ `BranchUniversalObject.getReferredBranchUniversalObject()`.
NATIVE_INIT_SESSION_FINISHED_EVENT_LINK_PROPERTIES: LinkProperties - If `params != null && params.has("~id")` this is `LinkProperties.getReferredLinkProperties()`.

This intent is now caught in the MainActivity of testbed_native_android and logged, like the NSNotification in testbed_native_ios.

Speaking of iOS, a BranchUniversalObject and BranchLinkProperties were also added to the RNBranchLinkOpenedNotification. This approach seems most convenient for the developer, allowing them the choice of using the BUO/[Branch]LinkProperties or the params.

Some other options would also be possible on both platforms, such as adding a callback to the RNBranch[Module].initSession method or adding a delegate/callback protocol/interface.

With the addition of the NSNotification/local broadcast intent, it is now at least possible for the native layer in each case to be notified when links are opened after react-native-branch is integrated into the app.

**Updated**

Added RNBranchInitListener, e.g.

```Java
        RNBranchModule.initSession(getIntent().getData(), this, new RNBranchInitListener() {
            @Override
            public void onInitFinished(Uri uri, BranchUniversalObject branchUniversalObject, LinkProperties linkProperties, BranchError error) {
```

This is more convenient than the local broadcast, which is still available.


android/src/main/java/io/branch/rnbranch/RNBranchModule.java
```diff
@@ -60,22 +65,55 @@ public void onInitFinished(JSONObject referringParams, BranchError error) {
...
+
+                    if (referringParams.has("~id")) {
+                        BranchUniversalObject branchUniversalObject = BranchUniversalObject.createInstance(referringParams);
+                        // BranchUniversalObject branchUniversalObject = BranchUniversalObject.getReferredBranchUniversalObject();
+                        if (branchUniversalObject != null) {
+                            broadcastIntent.putExtra(NATIVE_INIT_SESSION_FINISHED_EVENT_BRANCH_UNIVERSAL_OBJECT, branchUniversalObject);
+                        }
+
+                        // TODO: Can this be constructed with local state?
+                        LinkProperties linkProperties = LinkProperties.getReferredLinkProperties();
```
sojanpr: In this case the `LinkProperties` are created  with the `latest_referring_params` . So why we are  not using the 
`BranchUniversalObject.getReferredBranchUniversalObject()`  for creating the BranchUniversalObject?

examples/testbed_native_android/android/app/src/main/java/io/branch/testbed_native_android/MainActivity.java
```diff
@@ -78,7 +90,9 @@ protected void onDestroy() {
         if (mReactInstanceManager != null) {
             mReactInstanceManager.onHostDestroy(this);
         }
-    }
+
+        unsubscribeFromBranchLinks();
```
sojanpr: This seems like we developer need  to be aware  of adding this code  on destroy of every activity. Can SDK some how intercept the lifecycle methods and call the  `subscribeToBranchLinks` method  and `unsubscribeFromBranchLinks `

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
jdee: I suppose I usually prefer to use the callback context rather than stored state when handling asynchronous events. For example, there can be races. But a race is not much of a practical concern here, since these are user-initiated events. At any rate, I'd like things to be consistent. I can make them both use the latest_referring_params.

jdee:
@sojanpr I've added an `RNBranchInitListener` interface and an override that accepts this callback, which is more convenient than the local broadcast, which, however, is still available if anyone wants to use it. This is something I thought about doing anyway. I'll do something similar for iOS after this.

I've also changed it to use `BranchUniversalObject.getReferredBranchUniversalObject()` instead of constructing from the params in the callback.

aaustin:
@sojanpr Can you review this today?

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
sojanpr: I would suggest to use the method `getReferredBranchUniversalObject`. The reason is that in case Android SDK make any changes to the parsing of parameters ( thinking about adding new metadata with specific handling) you don't need to repeat the same thing here.

jdee: This has been done. I think you may be looking at an out-of-date commit.

examples/testbed_native_android/android/app/src/main/AndroidManifest.xml
```diff
@@ -26,6 +26,12 @@
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
             </intent-filter>
+            <intent-filter>
```
sojanpr: Please change to `<intent-filter android:autoVerify="true"> `  to support automatic app link verification through DAL

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
```diff
@@ -41,12 +46,18 @@
 
     private static JSONObject initSessionResult = null;
     private BroadcastReceiver mInitSessionEventReceiver = null;
+    private static RNBranchInitListener initListener = null;
```
sojanpr: This is  holding  a strong reference to the callback and prevent GC.  Do you think this can lead to a memory  leak. How about keeping this  initListener as a weak reference?

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
```diff
@@ -60,22 +71,57 @@ public void onInitFinished(JSONObject referringParams, BranchError error) {
...
+                    result.put(NATIVE_INIT_SESSION_FINISHED_EVENT_URI, uri != null ? uri.toString() : JSONObject.NULL);
                 } catch(JSONException ex) {
                     try {
                         result.put("error", "Failed to convert result to JSONObject: " + ex.getMessage());
                     } catch(JSONException k) {}
                 }
                 initSessionResult = result;
-                LocalBroadcastManager.getInstance(mmActivity).sendBroadcast(new Intent(NATIVE_INIT_SESSION_FINISHED_EVENT));
+
+                BranchUniversalObject branchUniversalObject =  BranchUniversalObject.getReferredBranchUniversalObject();
```
sojanpr: Can we use`Branch.BranchUniversalReferralInitListener()` instead of `Branch.BranchReferralInitListener()`. This callback is already giving `BranchUniversalObject` and `LinkProperties`
```branch.initSession(new Branch.BranchUniversalReferralInitListener() {
            @Override
            public void onInitFinished(BranchUniversalObject branchUniversalObject, LinkProperties linkProperties, BranchError error) ```

sojanpr:
Other than the above comments this looks good to me.

PR #169, opened 2017-04-18
Fix Android createLinkProperties

jchesne:
Hello,

I wanted to create a link with alias, unfortunelty this didn't worked. After a bit of research, its look like params : **alias**, **campaign**, **stage** were not properly modified via **setAlias**, **setCampaign**, **setStage** accordinaly in [createLinkProperties](https://github.com/BranchMetrics/react-native-branch-deep-linking/blob/master/android/src/main/java/io/branch/rnbranch/RNBranchModule.java#L321)

Version of Branch sdk : 0.14.12
Version of RN Branch : 1.1.1

### DIFF
--- a/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
+++ b/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
@@ -318,11 +318,11 @@ public class RNBranchModule extends ReactContextBaseJavaModule {
 
     public static LinkProperties createLinkProperties(ReadableMap linkPropertiesMap, @Nullable ReadableMap controlParams){
         LinkProperties linkProperties = new LinkProperties();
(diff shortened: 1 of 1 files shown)


jdee:
@jchesne Thanks for the PR. That looks like some copypasta gone wrong. I'm in the process of setting up native unit tests for exactly this sort of thing. Thanks for catching it before I did.

Branch would like to thank you for helping to improve this product. You can contact integrations@branch.io to receive swag if you so desire.

jdee:
@jchesne I have added your change to the `maintenance-1.1` branch. Would you be able to try that branch and verify? I intend to do a patch 1.1.2 release soon anyway, and this change will be there. (It will also be added to the maintenance-0.9 branch and released as a patch for RN < 0.40.)

PR #200, opened 2017-05-10
Create .npmignore

guysegal:
ignore native-tests library when using the npm package

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,1 @@
+native-tests/


guysegal:
closes https://github.com/BranchMetrics/react-native-branch-deep-linking/issues/199

rt2zz:
this will cause confusion as we *also* use the package.json files whitelist: https://github.com/BranchMetrics/react-native-branch-deep-linking/blob/master/package.json#L6-L12

moreover because native-tests is not part of files it should already be excluded. Is it possible you installed via git uri rather than npm release?

guysegal:
Hey Zack you are right, I didn't notice the 'files' section in the package.json, and I did used git url in my project. 

Once you'll publish new version of 2.0.0 I'll be able to use it properly.  

I'll remove the PR

rt2zz:
ya thats kind of rough, also a problem with babelrcs - makes its really hard to install from git forks :(

I recall someone had a package to help with this (run publish scripts for forked modules) but I cannot find it atm.

jdee:
It would be nice to be able to build an APK from a GitHub URL without the workaround in [this comment](https://github.com/BranchMetrics/react-native-branch-deep-linking/issues/194#issuecomment-300264629). For now, that's the only solution I've found that works. (The .npmignore fix didn't work when I tested it.)

I should have the next release out by the end of the day.

guysegal:
Btw - the npmignore works when using `npm i` how ever it's not working when using `yarn`

Issue #202, opened 2017-05-17
Expose debug-mode stuff from native Android lib, once working, in this lib


GeoffreyPlitt:
I've been in touch with David from Branch about how to simulate installs over and over again on Android, the way they do with IOS using the debug-flag trick.

According to David, the Android native lib approaches are the following, though bugs are currently being fixed and these don't work as of today:
- https://dev.branch.io/getting-started/integration-testing/guide/android/#use-debug-mode-to-simulate-fresh-installs
- enableSimulateInstalls() (In your Application class, beforethe getAutoInstance call, add: Branch.enableSimulateInstalls();)

I'm creating this issue to track the future work of exposing the Android part when the native libs are updated. Wanted an issue to subscribe to in order to make sure I know when this is in-progress / done. I'm not exactly sure how things get communicated between the native SDK maintainers and the maintainers of this lib.

jdee:
Hi @GeoffreyPlitt. Thanks for opening this issue. Unfortunately, a lot of this cannot be included directly in this module without a pretty large change. The problem is that the native layers are initialized long before the JavaScript loads, and the calls you refer to have to be made before the native layers initialize the Branch SDK, as you indicated. This is true for both iOS and Android.

However, there are numerous reasons that changing that might be a good idea. For example, there's also a request to set the Branch key(s) and other configuration from JS, which also has to be done before the native SDK is initialized.

There will at least be a performance impact. At the moment, the Branch SDK initialization, including a call to the API, occurs in parallel with loading the JS. Changing it will add some time to opening any initial link (when an app is launched from a link). This may also mean significant API changes (beyond just adding the required new methods).

I doubt this will make it into the production 2.0.0 release, but hopefully we can do something about it after that.

For now, of course, you can just make those modifications in the native code in your own app.

GeoffreyPlitt:
Gotcha, thanks!

GeoffreyPlitt:
Update: David just mentioned to me " I see that the fix for enableSimulateInstalls went live in the native Android repository this morning."

I do see what you mean about the complexity of exposing these settings in RN-land, how timing is affected by waiting until JS executes, and all of that.

But purely from a packaging perspective, if you can just update the native lib version so it's pulled in when I pull the RN lib, that would help, right? Is this ticket appropriate for that?

jdee:
Sure. I always update the native SDKs before a release. There will be another one soon, probably tomorrow.

jdee:
[...]

PR #204, opened 2017-05-19
Update Android SDK

jdee:
This updates the Android SDK to 2.8.0.

It also includes fixes for some bad behavior introduced in the last PR.


jdee:
This should provide immediate relief for #202.

PR #209, opened 2017-05-24
Enable apple ad attribution tracking in React Native Apps

ccubed510:
RNBranch doesn't call delayInitToCheckForSearchAds, so adding this as it is the wrapper we use in our react native app. 

Doesn't look like this is in the issue tracker, so wanted to add this functionality in.

#Test Plan
Ran NPM Test and fastlane iOS tests to verify no failures.

### DIFF
--- a/ios/RNBranch.m
+++ b/ios/RNBranch.m
@@ -70,6 +70,7 @@ RCT_EXPORT_MODULE();
 + (void)initSessionWithLaunchOptions:(NSDictionary *)launchOptions isReferrable:(BOOL)isReferrable {
     if (!branchInstance) {
         branchInstance = [Branch getInstance];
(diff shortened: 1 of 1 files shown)


jdee:
Thanks for the PR. Unfortunately, I don't think this is the right way to do this. For one thing, the `delayInitToCheckForSearchAds` is an optional call. We don't want to introduce that delay for all apps. This call has to be made before initializing the native session, so it cannot be exposed to JS. It has to be made in the native code in your own app.

There is now a convenient `[RNBranch delayInitToCheckForSearchAds]` method (see #210). You can add that just before the call to `[RNBranch initSessionWithLaunchOptions:isReferrable:]`. This will be in the next release.

I have some other ideas about ways to support these options without requiring native code modifications. For now, this should help.

Closing this PR.

PR #215, opened 2017-06-07
Add Branch API breaking changes in v1.1.0

Hitabis:
Complement docs with breaking changes in the definition of the createBranchUniversalObject function.


jdee:
Good catch. Thanks, @Hitabis.

PR #228, opened 2017-07-17
Breaking Change - Remove overridden createJSModules

christopherdro:
Required for React Native 0.47+

Reference commit: facebook/react-native@ce6fb33

### DIFF
--- a/android/src/main/java/io/branch/rnbranch/RNBranchPackage.java
+++ b/android/src/main/java/io/branch/rnbranch/RNBranchPackage.java
@@ -21,11 +21,6 @@ public class RNBranchPackage implements ReactPackage {
   }
 
   @Override
-  public List<Class<? extends JavaScriptModule>> createJSModules() {
-  	return Collections.emptyList();
-  }
-
-  @Override
   public List<ViewManager> createViewManagers(ReactApplicationContext reactContext) {
   	return Collections.emptyList();
   }


jdee:
Thank you for the PR. It's good to know that we need to do this for RN 0.47. A couple of thoughts:

- The Android unit tests failed in CI. I haven't looked closely to see where and why, but it might be related to this change. It's probably because the tests (devDependencies) use an earlier version of RN. You can see the failure for yourself using the `runAllTests` script in the repo or using `fastlane test_android`.
- A related issue: 2.0 is a fine place for a breaking change, but this means that we'll have to increase the peerDependency for RN to >= 0.47. This would mean that anyone using 1.x would be stuck until/unless they upgrade to 0.47. We still get reports from devs using RN 0.39. Version 2.0 is nearing release and is intended to replace 1.x entirely. I'm reluctant to increase the required version of RN in 2.0. But of course, without this change, you can't use RN 0.47.

We can either release 2.0 with RN >= 0.40 and then quickly release, say, 3.0 for 0.47 (major version increment required by semver rules for a breaking change). I'm not sure this is the best solution. We can probably find a trick either using automation (RN postlink hook) to modify the source depending on the RN version or maybe we can have two different versions of the module in the package for different versions of RN.

I've been working on another problem this week, but I hope to get the final beta release of 2.0.0 out soon, followed by a prod release. I'll look into the best way to support RN 0.47. Any suggestions (e.g. other SDKs solving this problem) are welcome.

jdee:
A possible approach to this issue:

1. Rename RNBranchModule (to RNBranchModuleBase, e.g.).
2. Make RNBranchModuleBase an abstract class.
3. Add two subclasses of RNBranchModuleBase: RNBranchModule (for >= 0.47) and RNBranchLegacyModule (for < 0.47).
4. In the docs, indicate that you should use RNBranchLegacyModule for < 0.47 and RNBranchModule otherwise.
5. Address the choice of class with automation when possible.

The base will have to be abstract in order to compile with 0.40, since it will be missing a method. I think something like this should work.

jdee:
This was addressed by #231. Removing the `@Override` is not a breaking change and works well with RN >= 0.47 and RN < 0.47.

PR #231, opened 2017-08-02
Remove createJSModules - RN 0.47 compatibility

radko93:
[`createJSModules` is now not required on Android](https://github.com/facebook/react-native/commit/ce6fb337a146e6f261f2afb564aa19363774a7a8) from RN 0.47. I actually can't build an app on RN 0.47 without removing this method. This is backwards compatible according to my tests (in comparison to #228).

### DIFF
--- a/android/src/main/java/io/branch/rnbranch/RNBranchPackage.java
+++ b/android/src/main/java/io/branch/rnbranch/RNBranchPackage.java
@@ -20,7 +20,7 @@ public class RNBranchPackage implements ReactPackage {
     return modules;
   }
 
-  // Deprecated RN 0.47
+  // Depreciated RN 0.47
   public List<Class<? extends JavaScriptModule>> createJSModules() {
   	return Collections.emptyList();
   }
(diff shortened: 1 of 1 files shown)


jdee:
Thank you for this pull request. See also #228, which addresses this issue. That PR suggests removing the method entirely.

In either case, my concern is breaking apps using RN < 0.47 or having to increase the peerDependency here for RN. I've been working on a solution to support both scenarios. The fix should get into NPM by the end of the week one way or another.

radko93:
This change is backawards compatible, I've tested this PR with RN 0.46.4: https://github.com/tinycreative/react-native-intercom/pull/116

jdee:
This was the right answer. The first PR was correctly labeled a breaking change and could not be supported with backwards compatibility without the kind of change in #232. I was also convinced from the `@Override` annotations that `ReactPackage` was an abstract class rather than an interface, which would also have required multiple flavors. This annotation is not required for an interface implementation, and in fact you can remove all the `@Override`s in the RNBranchPackage and everything builds fine without warnings with RN 0.43 or 0.47. If these are required (or recommended) for some other reason, Java and Android Studio don't seem to give you much help in detecting that.

At any rate, thanks for the contribution, @radko93. This will be released today.

PR #267, opened 2017-09-12
Fix openURL iOS native method compilation error

almirfilho:
I was following the setup documentation but running into a compilation failure with the native method `openURL`, which you have to put inside the _AppDelegate.m_ file, according to the README.

The error message:
__Semantic issue: Control may reach end of non-void function.__

Here's a printscreen of it:
screen shot 2017-09-12 at 5 37 27 pm

Just adding a `return NO;` at the end of the function fixed it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,7 @@ In AppDelegate.m
         // do other deep link routing for the Facebook SDK, Pinterest SDK, etc
         return YES;
     }
+    return NO;
 }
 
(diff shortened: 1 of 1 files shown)


jdee:
This is correct in the official docs: https://docs.branch.io/pages/apps/react-native/#initialize-branch. I'll update the README. Thanks for pointing it out.

PR #268, opened 2017-09-13
Allow loading rewards for different buckets

paulief:
This allows you to specify a bucket for which to load rewards. I saw this was implemented in the Cordova version so I added it here.

Linting and tests pass, and I checked that this works on  both platforms. Let me know if this looks good!

### DIFF
--- a/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
+++ b/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
@@ -585,9 +585,9 @@ public class RNBranchModule extends ReactContextBaseJavaModule {
     }
 
     @ReactMethod
-    public void loadRewards(Promise promise)
+    public void loadRewards(String bucket, Promise promise)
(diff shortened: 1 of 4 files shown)


android/src/main/java/io/branch/rnbranch/RNBranchModule.java
```diff
@@ -585,9 +585,9 @@ public void redeemRewards(int value, String bucket, Promise promise)
     }
 
     @ReactMethod
-    public void loadRewards(Promise promise)
+    public void loadRewards(String bucket, Promise promise)
     {
-        Branch.getInstance().loadRewards(new LoadRewardsListener(promise));
+        Branch.getInstance().loadRewards(new LoadRewardsListener(bucket, promise));
```
paulief: Android only has one `loadRewards` method

jdee:
Looks good. Thanks for the PR.

PR #280, opened 2017-10-10
fixing import to work correctly

blakedallen:
When installing via pods for ios, the documentation recommends using:
```objective-c
#import <react-native-branch/RNBranch.h> // at the top
```
To import the project, however this causes a file not found error. To correct:
```objective-c
#import "RNBranch.h"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ Modify your AppDelegate as follows:
 In AppDelegate.m
 
 ```objective-c
-#import <react-native-branch/RNBranch.h> // at the top
+#import "RNBranch.h" // at the top
 
 // Initialize the Branch Session at the top of existing application:didFinishLaunchingWithOptions:
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions


jdee:
Which version of RN are you using? If you're using < 0.40, then you may be stuck with version 0.9 of the SDK. This is getting harder for us to support, since, e.g., `react-native init` does not produce a project that is directly compatible with RN < 0.40.

For the mainstream use case here, the instructions are correct. The requirement to use RN >= 0.40 is mentioned in the docs. I believe everything is consistent. And you definitely do not use CocoaPods with 2.x, the current release of the SDK, which directly embeds both native SDKs via NPM.

It may be that a number of partners are still stuck using RN < 0.40, and we need some legacy docs for that purpose. But the change proposed here would not be correct. I'll revise the docs for the legacy case. Thanks for bringing this up.

blakedallen:
Version of react:
```
react-native-cli: 2.0.1
react-native: 0.46.4
```
react-native-branch version in node_modules/react-native-branch/package.json
```
"version": "2.1.0"
```

ryanvanderpol:
@jdee I am also receiving the same error and this seems to fix it.  I am using RN `0.48.4`.

However, I'm also getting some other errors when trying add the `openURL` method: `No visible @interface for 'RNBranch' declares the selector 'application:openURL:options:'` which may be related.

jdee:
Thanks for the feedback.

[The official docs](https://docs.branch.io/pages/apps/react-native/) should be considered canonical. The README may get out of synch from time to time, though we endeavor to keep them both accurate and current.

There is a section on common build problems: https://docs.branch.io/pages/apps/react-native/#common-build-problems

In general, you're going to get a much better result if you contact support instead of bringing issues to the repo. https://support.branch.io or integrations@branch.io. Branch has an excellent support team and will always make sure you get what you need.

But some thoughts on your issues:

@ryanvanderpol You need `[RNBranch.branch application:app openURL:url options:options]` (not `[RNBranch application:app openURL:url options:options]` (note: RNBranch.branch).

@blakedallen There must be something wrong in your project settings. You definitely cannot usually import headers from pods using `#import "Header.h"`. That's generally the wrong way to import framework/library headers anyway, CocoaPods notwithstanding. And you can try this:

```bash
react-native init NewProject
```

Then open the ios/NewProject.xcodeproj and have a look at the AppDelegate.m. Since RN 0.40, you need to import RN headers using `#import <React/RCTRootView.h>`, etc. If you change that to `#import "RCTRootView.h"`, it won't compile.

Before RN 0.40. React used to have you import headers that way. They had to modify a bunch of standard project settings to do it. They removed all of that in 0.40. Perhaps you updated from RN < 0.40 and did it manually, so the project settings were not converted. I've had good luck recently with the `react-native-git-upgrade` tool (`npm install -g react-native-git-upgrade; cd /path/to/myproject; react-native-git-upgrade`). I think it updates project settings. Perhaps it could improve the situation with your Xcode project, though I'm not sure.

If it's working for you with `#import "RNBranch.h"`, then I don't think there's any problem using it that way. But it might be worth looking into why things are working that way.

ryanvanderpol:
@jdee thanks for the response.  I don't mean to hijack this PR thread but perhaps I wasn't clear with my comment.  I copied and pasted everything directly from the docs and I am getting the error I mentioned above (and as seen in this screenshot):

image

Also note the same error as @blakedallen was experiencing as seen here:  

image

Changing this to `import "RNBranch.h"` makes that import error go away.

PR #299, opened 2017-12-08
Added requiresMainQueueSetup as required by v0.49+ of RN

UrbanChrisy:


jdee:
Thanks. This doesn't look like it will have any impact on previous versions. I updated most of the repo examples to 0.50.4 and didn't notice any problems without this change, but maybe I didn't look closely enough.

PR #327, opened 2018-03-14
Issue 322 async subscribe

sequoiaat:


src/index.js
```diff
@@ -5,7 +5,6 @@ const { RNBranch, RNBranchEventEmitter } = NativeModules
 import createBranchUniversalObject from './branchUniversalObject'
 import BranchEvent from './BranchEvent'
 
-export const DEFAULT_INIT_SESSION_TTL = 5000
```
sojanpr: @sequoiaat No sure why are we removing the DEFUALT_INIT_SESSION_TTL here. What would be the  time out  in this case if a request is not responding (Say an init session request)?

sequoiaat: DEFAULT_INIT_SESSION_TTL timeout was used to decide weather cached init session launch link should be returned when subscribe is called for the first time. This was leading to inconsistency when subscribe is called from a react-native async function, because at first install time, the time to get that called could take more than 5000ms. This was the root cause of the issue.

So we removed the timeout based decision to return the cache event and changed it to return on first call to subscribe, unless the used has chosen to skip cached events.

After this implementation, DEFAULT_INIT_SESSION_TTL became unused anywhere in the repo, and did not see any reason to keep it. Let us know if it is useful to keep for future use.

sequoiaat: @sojanpr  Can we commit the merge on this one? The failure is because the tests are not updated for the new logic. We will update them also

sequoiaat: We fixed the test cases also ? Shall we commit ?

sojanpr: LGTM 👍

PR #337, opened 2018-03-27
Clears cached latest session params when iOS application becomes inac…

sequoiaat:
…tive


sojanpr:
Can you please make sure init session is called again when application is active and  latest params are updated. If that works fine I am ok with this changes 👍

PR #344, opened 2018-04-27
Issue 314 int control param android causes invalid opg property

surabhi-sequoiaat:
Fixing issue : Integer values are read as double by default.

### DIFF
--- a/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
+++ b/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
@@ -1016,14 +1016,9 @@ public class RNBranchModule extends ReactContextBaseJavaModule {
                 if(readableMap.getDouble(key)%1==0){
                     return readableMap.getInt(key);
                 }
-                else
-                {
+                else{
                     return readableMap.getDouble(key);
                 }
-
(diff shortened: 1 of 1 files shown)


android/src/main/java/io/branch/rnbranch/RNBranchModule.java
```diff
@@ -1013,7 +1013,12 @@ private static Object getReadableMapObjectForKey(ReadableMap readableMap, String
             case Boolean:
                 return readableMap.getBoolean(key);
             case Number:
-                return readableMap.getDouble(key);
+                if(readableMap.getDouble(key)%1==0){
```
sojanpr: Please format the code with spacings

sojanpr:
@surabhi-sequoiaat Not sure what is the issue this PR is addressing. The values added to link params are converted to Strings before adding them. Are we seeing any issue when double values are added in the link param value?

sequoiaat:
@sojanpr The problem was that the integer values are getting converted to double before converting it in to String. Now we have changed it so that integer will get integer string and double will continue getting double string.

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
sequoiaat: i did the formatting, please review

sojanpr:
LGTM 👍

PR #355, opened 2018-07-09
Improve setup guide for Android

vomchik:
Avoid problem if some another activity also use `onNewIntent` method


aaaronlopez:
Thank you for catching! This has been updated. Closing this PR out.

PR #409, opened 2019-01-21
Android support for setting key from JS

jdee:
This is not quite finished or ready to merge, but this is the main part of the change for Android.


jdee:
Superseded.

PR #415, opened 2019-02-08
Reset user session in all cases when isFromSelf

c100k:
If you open a link `scheme://` without `link_click_id` while the application is running, it does not pass it to the JS `subscribe` method because the user session is not reset.

However, if you put the app in background and then foreground, it opens the link.

If we reset the user session in any case (`link_click_id` or not) if `isFromSelf == true`, it successfully call the JS `subscribe` method.


jdee:
This is a change to the native iOS SDK. There is a copy in the RN module, but it is automatically updated, and this is very much the wrong place to address this, I'm afraid. If we merge this, it will be overwritten the next time we automatically update the embedded native SDK. There is an open ticket for the iOS team, and they are addressing this. Once that is merged and released in the native iOS SDK, we'll release an update of the RN SDK with the updated iOS component.

c100k:
@jdee I understand. Can you link the issue in the native repository so that we can follow it ? Thanks.

Issue #401, opened 2018-12-29
pod 'Branch' vs pod 'Branch-SDK'


pcooney10:
In the setup instructions, we are advised to replace any references to `pod 'Branch'` with `pod 'Branch-SDK', path: "../node_modules/react-native-branch/ios"`.
**What is the difference between the two?**

For context, I am trying to set up Branch with Segment, and the `'Segment-Branch'` pod has the `'Branch'` pod as a dependency: [Segment-Branch.podspec#L27](https://github.com/BranchMetrics/Segment-Branch-iOS/blob/master/Segment-Branch.podspec#L27).
I am getting errors like `duplicate interface definition for class 'BNCPreferenceHelper'` and `property has a previous declaration`.

Would I be able to depend on `pod 'Branch'` instead of `pod 'Branch-SDK', path: "../node_modules/react-native-branch/ios"`? Or do I need to create a fork of `'Segment-Branch'` that doesn't include `Branch` as a dependent?

CC'ing #379 as it's related

sequoiaat:
@pcooney10 - if we understand the question correctly  then yes   pod 'Branch' references have to be changed with pod 'Branch-SDK'. However we think you are facing an issue because you are using Segment-branch inside this which has a pod named branch . Could you try to create a fork of segment-branch and see if it solves the issue else we will try to replicate it .

pcooney10:
Hi @sequoiaat, thank you for the response.

Outside of pod 'Branch' references needing to be changed to pod 'Branch-SDK', I guess my question was what is the difference between the two? Would there be a lot of difficulty involved with having the react native implementation for iOS to reference pod 'Branch'?  I could be wrong, but it seems like importing `#import <Branch/Branch.h>` works pretty much the same as `#import <react-native-branch/RNBranch.h>`, and I can reference both `RNBranch` and `Branch` interchangeably. 

[...]

PR #422, opened 2019-02-28
Segment fix

jdee:
Addresses #401.

This adds a `react-native-branch-segment` pod that addresses a dependency conflict.

The pods here are only intended for use with hybrid apps: native iOS apps written in Swift or Obj-C that include one or more RN components. If you're using `react-native link` (or `rnpm link`) to build your app, the Branch native iOS SDK is embedded in the `react-native-branch` project.

To use this pod in an app using the Branch Segment integration, add this to your Podfile:

```Ruby
pod "react-native-branch-segment", path: "node_modules/react-native-branch" # or wherever your node_modules is
```

Instead of

```Ruby
pod "react-native-branch", path: "node_modules/react-native-branch"
pod "Branch-SDK", path: "node_modules/react-native-branch/ios"
```

Be sure you're not including the Branch-SDK pod from node_modules. This may not be the last word, but it fixes the conflict and may be considered an officially-supported solution.

### DIFF
--- a/react-native-branch-segment.podspec
+++ b/react-native-branch-segment.podspec
@@ -1,5 +1,8 @@
 require 'json'
 
+# Podspec specifically for use in hybrid apps with the Branch Segment integration.
+# Not for use in pure RN apps.
+
 # expect package.json in current dir
 package_json_filename = File.expand_path("./package.json", __dir__)
 
@@ -20,7 +23,7 @@ Pod::Spec.new do |s|
   s.platform     = :ios, "7.0"
   s.source       = { spec['repository']['type'].to_sym => spec['repository']['url'].sub(/^[a-z]+\+/, '') }
   s.source_files = [ "ios/*.h", "ios/*.m"]
-  s.header_dir   = "react_native_branch"
+  s.header_dir   = "react-native-branch"
   s.dependency 'Branch'
   s.dependency 'React' # to ensure the correct build order
 end
--- a/react-native-branch-segment.podspec
+++ b/react-native-branch-segment.podspec
@@ -1,0 +1,26 @@
+require 'json'
+
+# expect package.json in current dir
+package_json_filename = File.expand_path("./package.json", __dir__)
+
+# load the spec from package.json
+spec = JSON.load(File.read(package_json_filename))
+
+Pod::Spec.new do |s|
+  s.name             = "#{spec['name']}-segment"
+  s.version          = spec['version']
+  s.summary          = spec['description']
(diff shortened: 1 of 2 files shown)


react-native-branch-segment.podspec
```diff
@@ -0,0 +1,29 @@
...
+  s.authors      = {
+                     'rt2zz' => 'zack@root-two.com',
+                     'Jimmy Dee' => 'jgvdthree@gmail.com'
+                   }
+  s.license      = spec['license']
+  s.homepage     = spec['homepage']
+  s.platform     = :ios, "7.0"
+  s.source       = { spec['repository']['type'].to_sym => spec['repository']['url'].sub(/^[a-z]+\+/, '') }
+  s.source_files = [ "ios/*.h", "ios/*.m"]
+  s.header_dir   = "react-native-branch"
```
jdee: This ensures that imports don't have to change, if you're importing react-native-branch in your native code. This isn't actually necessary unless you're using `branch.subscribe` in your RN (JS) code. If you're routing links in native code, which seems the most likely thing, you can just use the Branch native iOS SDK directly. This is how webview_example_native_ios is set up. The testbed_native_ios app uses the RNBranch native SDK as an example and verifies that this change builds correctly.

react-native-branch-segment.podspec
```diff
@@ -0,0 +1,29 @@
...
+  s.dependency 'Branch'
```
jdee: Corrected dependency

react-native-branch-segment.podspec
```diff
@@ -0,0 +1,29 @@
...
+# Not for use in pure RN apps.
+
+# expect package.json in current dir
+package_json_filename = File.expand_path("./package.json", __dir__)
+
+# load the spec from package.json
+spec = JSON.load(File.read(package_json_filename))
+
+Pod::Spec.new do |s|
+  s.name             = "#{spec['name']}-segment"
```
jdee: Pod name change

echo-branch:
Code looks good!  I'll try to test it myself either later tonight or tomorrow.

jdee:
To test, testbed_native_ios is the best option, since it uses the RNBranch native code. Just change the Podfile there as described above. Then in that directory (examples/testbed_native_ios) run:

```bash
yarn
bundle check || bundle install
bundle exec pod install
```

Then you can build the workspace.

echo-branch:
Works in test app!  I see the updated podspec pulls in a single copy of branch.

PR #426, opened 2019-04-04
Back to CocoaPods

jdee:
This change removes the embedded Branch iOS SDK from this package. The iOS SDK must now be taken from CocoaPods. To convert a pure React Native app built with react-native link:

- Add a Podfile to your project under the ios subdirectory if you don't already have one. It should have these contents:
    ```Ruby
    platform :ios, "9.0"
    use_frameworks!
    pod "Branch", "0.26.0"
    target "MyApp"
    ```
    Use the name of your app target instead of MyApp. The `use_frameworks!` directive is optional but recommended.
    Your deploy target in the `platform` directive may vary.
- If you already have a Podfile, add:
    ```Ruby
    pod "Branch", "0.26.0"
    ```
- After modifying your Podfile, run `pod install --repo-update`. [TODO: Should this be `pod update` in some cases?]
- If you have just added the Podfile to your project, a new workspace will be created called `MyApp.xcworkspace` (assuming `MyApp.xcodeproj`). From now on, open the workspace instead of the project to build your app.

If you have a native iOS app that includes a RN component using this SDK, remove `pod "Branch-SDK"` from your Podfile and run `pod install`.

If you are using the `react-native-branch-segment` pod, replace it with `react-native-branch` in your Podfile and run `pod install`.

Notes:
- The `use_frameworks!` directive is recommended but optional. The `RNBranch.xcodeproj` has both a Framework Search Path and a Header Search Path to pick up Branch imports. We don't want to impose this choice on Partners. We have to support both options.
- This has been tested with Xcode 10.2.
- ~~The webview_example_native_ios app is throwing a runtime exception from JS. This will be fixed.~~
- All example apps udpated to RN 0.59.3 or 0.59.4 and React 16.8.3. ~~Both testbed_native_ios and webview_example_native_ios have been updated to RN 0.59.3, React 16.8.3. The pure RN apps and testbed_native_android should also be updated.~~
- The Branch SDK is required in the `react-native-branch.podspec` using ~~`~> 0.26.0`. This pins the minor version but allows patch updates. Is this best?~~ `0.26.0`. This pins to a specific release.
- All iOS example apps have been updated. Some have `use_frameworks!`. Some do not.
- The `webview_tutorial` needs to be updated (docs only).
- The Fastlane automation that updated the embedded SDK needs to be updated to modify the podspec and the ~~MINIMUM_BRANCH_SDK_VERSION~~ REQUIRED_BRANCH_SDK_VERSION. This is not much work.
- Apologies to @echo-branch for the huge PR. Most of the changes are automated from CocoaPods. I'll highlight key changes with line comments.
- This package will perform runtime validation of the native iOS SDK version. This is the only way to ensure the correct version is present with pure RN apps. If the minimum SDK version is not met, an error is logged. This is visible as an overlay in the app in dev mode at least. The current minimum is 0.26.0. This screenshot was generated by increasing the required minimum to 0.26.1, which does not yet exist in CocoaPods.
    Simulator Screen Shot - iPhone X - 2019-04-03 at 18 17 33


react-native-branch.podspec
```diff
@@ -20,6 +20,6 @@ Pod::Spec.new do |s|
   s.platform     = :ios, "7.0"
   s.source       = { spec['repository']['type'].to_sym => spec['repository']['url'].sub(/^[a-z]+\+/, '') }
   s.source_files = [ "ios/*.h", "ios/*.m"]
-  s.dependency 'Branch-SDK'
+  s.dependency 'Branch', '~> 0.26.0'
```
jdee: Here we pin in the podspec.

ios/RNBranch.xcodeproj/project.pbxproj
```diff
@@ -370,253 +62,36 @@
...
-		F3744CDD5733CE0539D480D9 /* BNCApplication.m */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.c.objc; path = BNCApplication.m; sourceTree = "<group>"; };
-		F905420D27F1836548D7618E /* BNCThreads.m */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.c.objc; path = BNCThreads.m; sourceTree = "<group>"; };
 /* End PBXFileReference section */
 
 /* Begin PBXFrameworksBuildPhase section */
 		B3A3CC331C5B0A070016AC52 /* Frameworks */ = {
 			isa = PBXFrameworksBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
+				7BA40F94225589C900EC6B4D /* Branch.framework in Frameworks */,
```
jdee: Branch.framework was added as an optional dependency in the Link Binary With Libraries Build Phase to ensure that it is built before RNBranch.xcodeproj, or imports fail. Making it optional allows it also to work when `use_frameworks!` is not present.

ios/RNBranch.m
```diff
@@ -28,6 +29,22 @@
...
+    int major = components[0].intValue;
+    int minor = components[1].intValue;
+    int patch = components[2].intValue;
+
+    NSArray<NSString *> *required = [MINIMUM_REQUIRED_BRANCH_SDK componentsSeparatedByString:@"."];
+    int requiredMajor = required[0].intValue;
+    int requiredMinor = required[1].intValue;
+    int requiredPatch = required[2].intValue;
+
+    return major > requiredMajor || (major == requiredMajor && (minor > requiredMinor || (minor == requiredMinor && patch >= requiredPatch)));
```
jdee: I'm not sure if there is a more standard way in Obj-C to compare version numbers. This was straightforward but off the cuff.

ios/RNBranch.m
```diff
@@ -75,6 +92,10 @@ + (BOOL)requiresMainQueueSetup {
 + (void)setupBranchInstance:(Branch *)instance
 {
     RCTLogInfo(@"Initializing Branch SDK v. %@", BNC_SDK_VERSION);
+    if (!branchSdkVersionValid()) {
+        RCTLogError(@"Please use at least v. %@ of Branch. In your Podfile: pod 'Branch', '~> %@'. Then pod install.", MINIMUM_REQUIRED_BRANCH_SDK, MINIMUM_REQUIRED_BRANCH_SDK);
```
jdee: RCTLogError generates the overlay at least in dev mode and renders the app useless. This can be reduced to a warning, but perhaps because this happens before JS loads, the warning did not show up in the app. Warning overlays can be dismissed and temporarily ignored.

echo-branch:
Looks good!  Lets update docs to say just use yarn.

PR #427, opened 2019-04-04
Update instructions to upgrade to v3

sonicdoe:
`implementation fileTree(dir: "libs", include: ["*.jar"])` is included in React Native’s `android/app/build.gradle` [by default](https://github.com/facebook/react-native/blob/3b91a7ec234b9aebfe11bb1c94de7b41444b57f2/template/android/app/build.gradle#L141). Removing it might affect other packages and since the `libs` directory was removed in 1d2925b1fa954bfe6088deb75958672533b7af1f anyway, it should not be necessary to remove this line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,12 +99,6 @@ module, starting with version 3.0.0.
 
 Open the `android/app/build.gradle` file in your project.
 
-Remove this line:
-
-```gradle
-implementation fileTree(dir: 'libs', include: ['*.jar'])
-```
-
 Add this line:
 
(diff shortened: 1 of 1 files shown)


jdee:
Thanks for the PR, and sorry for the delay. I think we've corrected this in the main README by now. We simply say we no longer require it. Anyway, this PR now has conflicts. If you think the wording still needs to be improved, please feel free to submit a new PR or even just open an issue and ask us to correct it.

PR #442, opened 2019-05-02
Fix examples

jdee:
Converted all examples that use a RN WebView to use react-native-webview to remove a deprecation warning. This exposed some important issues.

First, I was wrong about compatibility between the React pod from node_modules and react-native link. react-native link will work in a project using the React pod, as long as the Podfile is under the ios subdirectory (unlike webview_example_native_ios and testbed_native_ios). However, instead of adding a static library under the Libraries group in the Xcode project, react-native link will add a pod from node_modules to the Podfile, e.g. `pod 'react-native-webview', path: '../node_modules/react-native-webview'`. However, just like the react-native-branch pod, other pods from node_modules, including react-native-webview, will require the React pod, which was discontinued in CocoaPods. If you don't also set up the React pod with Yoga and Folly and remove the Libraries group if you have one from the Xcode project to avoid conflicts, pod install will quietly install version 0.11 of the React pod, the last prod release to CocoaPods.

In essence, if it detects a Podfile, react-native link assumes you are using the React pod from node_modules. If you are not (as with, e.g. webview_example), react-native link cannot be used or at least must be manually assisted. I converted browser_example and webview_example manually without react-native link. In other words, once you update to react-native-branch 3.0.0, react-native link will be broken for your project until you convert to use the React pod. This is actually an improvement, since it puts all native dependencies under control of CocoaPods and will simplify matters with regard to the native Branch SDK. We probably need either to document the steps to make this transition or find them documented elsewhere. We are forcing our users to convert to a different configuration. However, I suspect it's a more common configuration than I previously thought. This configuration is illustrated in #441. I originally thought this was an outlying use case, but this is probably going to be the main configuration for most RN apps that use react-native-branch.

Further, react-native link will experience a fatal error when trying to modify the Podfile if you don't use the target-block style, which is a bit more verbose than what we've recommended in the README. So all the example Podfiles and the snippet in the README have been modified to use the target-block style.

Another modification was made in native iOS code in testbed_native_ios to eliminate a warning in the app.

I've been working on eliminating the deprecated RN ListView in favor of FlatList, but I haven't finished that. It will be a separate PR.


jdee:
Based on this, I think we have to recommend that apps using react-native-branch convert to using the React pod from node_modules if they aren't already. This amounts to changing your Podfile from:

```Ruby
target 'MyApp' do
  pod 'Branch', '0.27.0'
end
```

to

```Ruby
target 'MyApp' do
  # React pods and subspecs required by most apps
  pod 'React',
    path: '../node_modules/react-native',
    subspecs: %w[
      Core
      CxxBridge
      DevSupport
      RCTAnimation
      RCTImage
      RCTText
      RCTNetwork
      RCTWebSocket
    ]
  pod 'yoga', path: '../node_modules/react-native/ReactCommon/yoga'
  pod 'Folly', podspec: '../node_modules/react-native/third-party-podspecs/Folly.podspec'

  # Import react-native-branch pod from node_modules
  pod 'react-native-branch', path: '../node_modules/react-native-branch'
end
```

And then remove the entire Libraries group from your Xcode project. (Choose "Remove References" rather than "Move to Trash." The same files are used by the React pod.)

Screen Shot 2019-05-02 at 10 01 55 AM

Screen Shot 2019-05-02 at 10 02 08 AM

I also manually added "Start Packager" and "Bundle React Native code and images" Run Script build phases to each project. These are present in the React.xcodeproj that is included when you run `react-native init`. I copied these out of a fresh reference project generated by `react-native init` from the React.codeproj in the Libraries group. The actual scripts have changed over time. It's probably best to copy from a fresh project, but the exact script for each build phase is listed as well.

Screen Shot 2019-05-02 at 10 05 52 AM

**Start Packager** run script build phase:
```bash
export RCT_METRO_PORT="${RCT_METRO_PORT:=8081}"
echo "export RCT_METRO_PORT=${RCT_METRO_PORT}" > "${SRCROOT}/../node_modules/react-native/scripts/.packager.env"
if [ -z "${RCT_NO_LAUNCH_PACKAGER+xxx}" ] ; then
  if nc -w 5 -z localhost ${RCT_METRO_PORT} ; then
    if ! curl -s "http://localhost:${RCT_METRO_PORT}/status" | grep -q "packager-status:running" ; then
      echo "Port ${RCT_METRO_PORT} already in use, packager is either not running or not running 
correctly"
      exit 2
    fi
  else
    open "$SRCROOT/../node_modules/react-native/scripts/launchPackager.command" || echo "Can't start packager automatically"
  fi
fi
```

Screen Shot 2019-05-02 at 10 06 11 AM

**Bundle React Native code and images** run script build phase:
```bash
export NODE_BINARY=node
../node_modules/react-native/scripts/react-native-xcode.sh
```

Doing this means that `react-native link` will continue to work correctly, and you don't have to specify and update a version requirement for the Branch SDK in your own Podfile in order to get the correct release.

I'd prefer to find some automation or standard RN documentation for doing this, but failing that, we'll have to include these instructions in our docs.

PR #447, opened 2019-05-03
Added detailed instructions for converting to use the React pod.

jdee:


jdee:
To test this with a new project:

```bash
react-native init TestApp
cd !$
yarn add react-native-webview
react-native link !$
```

Now start with `react-native unlink react-native-webview` in Step 1. This may not be necessary if you remove the Libraries group before running `react-native link` for each dependency again.

PR #459, opened 2019-05-23
RNBranch import (proposal)

### LINKED ISSUE #450: Simplify Obj-C import
Currently it is necessary to use two different styles of import in Objective-C depending whether you have `use_frameworks!` in your Podfile:

With `use_frameworks!`:
```Obj-C
@import react_native_branch;
```

Without `use_frameworks!`:
```Obj-C
#import <react-native-branch/RNBranch.h>
```

It is possible to simplify this with one line in the podspec: `s.header_dir = 'RNBranch'`. This allows `#import <RNBranch/RNBranch.h>`. It also changes the name of the module so that you `@import RNBranch;`. Clang (or some component of the build chain) will automatically convert things like `#import <RNBranch/RNBranch.h>` to `@import RNBranch;` when possible. This can reduce build times and have other benefits. In particular, that style of import always works no matter the build settings. We could recommend a single import instead of two.

I tried to solve this by adding a `react-native-branch.h` header hoping that `#import <react-native-branch/react-native-branch.h>` would become `@import react_native_branch;`. But this doesn't work, probably because of the hyphens and underscores.

To change the header_dir in the podspec would be a breaking change now that 3.0.0 is released to prod, so would basically force 4.0.0. This could be done as a non-breaking change by introducing a second podspec, but then we'd have to have special instructions for that podspec, which is probably worse than explaining two imports.

This is a minor issue, but seems worth noting that it can be improved.

jdee:
Fix #450 

**THIS IS A BREAKING CHANGE.**

However, CocoaPods 1.7 already breaks any build using the react-native-branch and React pods. I tried to fix it by adding `s.header_dir = 'react-native-branch'` to the podspec, but that is just the default setting, and it doesn't change the behavior. I may be able to find a simpler workaround.

But we also have the issue of multiple Obj-C imports depending whether modules are in use. Ordinarily clang will convert something like `#import <RNBranch/RNBranch.h>` to `@import RNBranch;` when it can. I tried adding a `react-native-branch.h` so you would use `#import <react-native-branch/react-native-branch.h>`, but clang isn't smart enough to turn that into `@import react_native_branch;` because of the hyphens.

In addition, the issue is not limited to the podspec. The RNBranch.xcodeproj also has to change. This means it will be fairly hard to support both module names/header paths.

This change would mean:

Swift:

`import react_native_branch` -> `import RNBranch`

Obj-C:

`#import <react-native-branch/RNBranch.h>` -> `#import <RNBranch/RNBranch.h>`
`#import <react_native_branch/RNBranch.h>` -> `#import <RNBranch/RNBranch.h>`
`@import react_native_branch;` -> `#import <RNBranch/RNBranch.h>` (optionally, or `@import RNBranch;`)

This would simplify documentation and support and make iOS projects less brittle against CP versions and configurations.

Unfortunately, semantic versioning would require this to be 4.0. This should have been rolled into 3.0, which was just released. But arguably this is a bug fix for the CP 1.7 issue. I've split the difference in the second commit and made it a minor version increment.

I don't know if this is a good idea given that it's a breaking change. All feedback welcome.


jdee:
It's worth noting that anyone not using the React pod (using the CocoaPods half-solution of just having `pod 'Branch', '0.27.0'` in a Podfile or using Carthage) is not affected by the CP 1.7 issue. So this breaking change just inconveniences them.

jdee:
Given that this will be fixed in CocoaPods 1.7.1, this solution becomes less attractive.

PR #497, opened 2019-09-10
Update React-Native-Branch v3 to support Branch ios v0.28.1

matthova:
Fixing error:
```
No visible @interface for 'Branch' declares the selector 'setAppleSearchAdsDebugMode'
```

by calling static method instead of instance.

This patch is necessary to build with ios branch version 0.28.1

### DIFF
--- a/ios/RNBranch.m
+++ b/ios/RNBranch.m
@@ -90,7 +90,7 @@ RCT_EXPORT_MODULE();
         [instance delayInitToCheckForSearchAds];
     }
     if (config.appleSearchAdsDebugMode) {
-        [instance setAppleSearchAdsDebugMode];
+        [RNBranch setAppleSearchAdsDebugMode];
     }
 }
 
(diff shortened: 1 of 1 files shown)


matthova:
@radko93 good catch. I've updated that and all files (excluding the examples)

matthova:
I should add, in the app where I'm using this library, we can't yet upgrade to react-native 60 so we are using react-native-branch v3

radko93:
Ok then we also have to update v4 branch, I might do it but probably next week.

rnagella:
@radko93 I am following this thread and just wondering if there is anything blocking you to tackle this. Thanks for the help.

radko93:
@rnagella nope but I didn't have time yet. It's easy to upgrade but not that fun to test things.

matthova:
@echo-branch any thoughts on this PR?
Do we want to bump it straight to 29.3?

echo-branch:
@matthova I haven't had a chance to get caught up on this.  My last couple sprints have been completely devoted to iOS.

We should be on-boarding a new maintainer for React Native and Cordova soon.  So hopefully we can bring our response times back to reasonable.

PR #499, opened 2019-09-11
Expose set request metadata to JS

ssg-luke:
Found its quite important to expose this function to the JS. In the case of Mixpanel integration the distinct ID may change while the app is running and you need to call this again in order to update Branch. Its actually mentioned in your documentation here https://docs.branch.io/integrations/mixpanel/#pass-mixpanel-distinct-id ^_^


echo-branch:
Although this PR would work for the Mixpanel situation, the method would not work in other cases.  In particular early in the app lifecycle before the JS code has a chance to run.  Considering our options on this, we may just merge this and document it's limitations.

PR #503, opened 2019-09-23
INTENG-6825 android push handling

echo-branch:
Andy added a reInit session method to handle foreground pushes and programmatic linking.  I added the same method to the RNBranch class.  

In the MainActivity onNewIntent, adding RNBranchModule.reInitSession(this) should allow the app to handle programmatic links and foreground pushes.


echo-branch:
Going to just merge this and test on stage.  :/

PR #508, opened 2019-09-30
Fix minimum native iOS SDK requirement

radko93:
Without this fix, it's not possible to use this library.


echo-branch:
Good catch!  Totally forgot to commit that last one.

PR #520, opened 2019-11-27
Make it explicit that manual step is required when using branch.json with RN > 0.60

mieszko4:
When installing fresh with RN > 0.60 it is unclear that manual step is required when using branch.json

### DIFF
--- a/docs/branch.json.md
+++ b/docs/branch.json.md
@@ -38,6 +38,8 @@ react-native link react-native-branch
 
 ### Manual integration without react-native link
 
+**NOTE: This is required for RN > 0.60**
+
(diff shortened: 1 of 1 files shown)


mieszko4:
Also, I think the manual step should say to:
1. put `branch.json` into root (where `package.json` is)
2. link it using symlink `cd android/app/src/main/assets/ && ln -s ../../../../../branch.json` for android (instead of copying)
3. for ios it seems good but is unclear that the file should be located in root project and linked from there.

We should have one source of truth (one configuration file)

gdeluna-branch (approved):

PR #542, opened 2020-02-24
SDK-854 only call back on successful completion.  ignore cancel

echo-branch:
See ticket for details.


jdee:
:+1: Looks good, but isn't there the potential for the same thing on Android?

echo-branch:
Potentially.  Would have to check the corresponding android code.

echo-branch:
Checked the android code.  I don't think android calls back with status information on cancel.

PR #550, opened 2020-03-09
3.x examples

jdee:
Updated all examples in the maintenance-3.x branch to use react-native-branch@3.1.2, which uses iOS 0.31.3 & Android 4.3.2 (Play Install Referrer). These updates avoid androidx for some other RN dependencies (react-native-gesture-handler, react-native-screens).

Also updated all apps to RN 0.59.9 to get past a consistent failure in all apps using Xcode 11. This has nothing to do with react-native-branch. The peer dependency for 0.59.9 is React 16.8.3.

All these examples now run on both iOS and Android.

Note all changes are dependencies, not code.

### DIFF
--- a/examples/browser_example/ios/Podfile.lock
+++ b/examples/browser_example/ios/Podfile.lock
@@ -9,59 +9,59 @@ PODS:
     - DoubleConversion
     - glog
   - glog (0.3.4)
-  - React (0.59.6):
-    - React/Core (= 0.59.6)
+  - React (0.59.9):
+    - React/Core (= 0.59.9)
   - react-native-branch (3.1.2):
     - Branch (= 0.31.3)
     - React
   - react-native-webview (5.12.1):
     - React
-  - React/Core (0.59.6):
-    - yoga (= 0.59.6.React)
-  - React/CxxBridge (0.59.6):
+  - React/Core (0.59.9):
+    - yoga (= 0.59.9.React)
+  - React/CxxBridge (0.59.9):
     - Folly (= 2018.10.22.00)
     - React/Core
     - React/cxxreact
     - React/jsiexecutor
-  - React/cxxreact (0.59.6):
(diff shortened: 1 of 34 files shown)


bklastaitis-branch:
@jdee I haven't looked at the config changes yet but have a couple observations.

- I was able to run `testbed_native_android` (on emulator API 28) but I keep seeing the overlay permission screen on launch (see below). Why does it happen and can/should we fix it?
- `redeemRewards` keeps throwing an error `Trouble redeeming rewards. Unable to redeem rewards. Please make sure you have credits available to redeem.` (this is after clicking loadRewards)
- Install_referrer entries should be removed from the manifest files
- After clicking around, closing and reopening the app I can fairly regularly reproduce the NPE we have seen from a few clients now (https://gist.github.com/bklastaitis-branch/ce2cfc504fc0bed4347ccf7e77f5b429). I believe you thought it was related to the install referrer lib, so I tried adding that dependency + removing the manifest entry but still got the error. Let me know if you can't reproduce this and we can share screens, see what's happening.

Screenshot_1583789365

jdee:
Thanks for testing, @bklastaitis-branch. I'm afraid testbed_native_android is one I would least expect to work well. That is an old app, and the RN docs used to tell you to put that in. It's only for the dev mode screen that RN automatically shows in a pure RN app. Did you try just webview_example or testbed_simple on Android?

I couldn't find the InstallListener in any of the manifests at first, but I see it now in some of them. I'll remove it.

I'll synch up with you separately about the crash you're seeing. Thanks.

bklastaitis-branch:
Synched up in private, figured out the permissions overlay thingy (expected), reproduced the bug in testbed_native_android and, because it was easy to reproduce, Jimmy said he will look into it and and add the fix to this PR, so I'll wait for that to complete the review.

PR #551, opened 2020-03-10
[INTENG-8649] Facebook deep linking class

jdee:
Added support for an option in branch.json, `enableFacebookLinkCheck`, e.g.:

```json
{
  "branchKey": "key_live_xxxx",
  "enableFacebookLinkCheck": true
}
```

This is used to call the `enableFacebookAppLinkCheck()` method on Android. On iOS, it uses reflection to find the `FBSDKAppLinkUtility` class to pass to `registerFacebookDeepLinkingClass:` on the Branch instance.

This provides an alternative to native code for supporting this option for RN users.


android/src/main/java/io/branch/rnbranch/RNBranchModule.java
```diff
@@ -687,6 +687,8 @@ private static Branch setupBranch(Context context) {
 
             if (mUseDebug || config.getDebugMode()) branch.setDebug();
 
+            if (config.getEnableFacebookLinkCheck()) branch.enableFacebookLinkCheck();
```
echo-branch: I think the method name should be enableFacebookAppLinkCheck.

leotm:
Hey @jdee, should this resolve https://github.com/facebook/react-native-fbsdk/issues/648#issuecomment-611253282?

jdee:
@leotm I would think so. In particular if you have `enableFacebookLinkCheck: true` in branch.json, and you don't see this https://github.com/BranchMetrics/react-native-branch-deep-linking-attribution/blob/3644288535034ea9712e8ce9eb4ac6440a39ecf1/ios/RNBranch.m#L91, then I would expect it to be set up correctly. This is where the actual call to the Facebook iOS SDK happens in the Branch iOS SDK: https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/blob/master/Branch-SDK/Branch-SDK/BNCFacebookAppLinks.m#L67.

leotm:
I am unsure if I am combining https://github.com/facebook/react-native-fbsdk

```c
- (BOOL)application:(UIApplication *)app
            openURL:(NSURL *)url
            options:(NSDictionary<UIApplicationOpenURLOptionsKey,id> *)options
{
  if ([[FBSDKApplicationDelegate sharedInstance] application:app openURL:url options:options]) {
    return YES;
  }

  if ([RCTLinkingManager application:app openURL:url options:options]) {
    return YES;
  }

  return NO;
}
```

and https://help.branch.io/developers-hub/docs/react-native

```c
#import <RNBranch/RNBranch.h> // at the top
// #import <react-native-branch/RNBranch.h> // with 3.x

// Initialize the Branch Session at the top of existing application:didFinishLaunchingWithOptions:
- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
    // Uncomment this line to use the test key instead of the live one.
    // [RNBranch useTestInstance];
    [RNBranch initSessionWithLaunchOptions:launchOptions isReferrable:YES]; // <-- add this
    NSURL *jsCodeLocation;
    //...
}

- (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDictionary<UIApplicationOpenURLOptionsKey,id> *)options {
    if ([RNBranch application:app openURL:url options:options])  {
        // do other deep link routing for the Facebook SDK, Pinterest SDK, etc
    }
    return YES;
}

- (BOOL)application:(UIApplication *)application continueUserActivity:(NSUserActivity *)userActivity restorationHandler:(void (^)(NSArray *restorableObjects))restorationHandler {
    return [RNBranch continueUserActivity:userActivity];
}
```

... correctly.

Does this like the correct result for `AppDelegate.m`?
And can you see any other libraries/settings possibly interfering?

```c
#import "AppDelegate.h"

#import <Firebase.h>
#import "RNFirebaseNotifications.h"
#import "RNFirebaseMessaging.h"

#import <AppCenterReactNative.h>

#import <CodePush/CodePush.h>

#import <React/RCTBundleURLProvider.h>
#import <React/RCTRootView.h>

#import <RNBranch/RNBranch.h>

#import <Branch/Branch.h>

#import <FBSDKCoreKit/FBSDKCoreKit.h>

@implementation AppDelegate

- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
  [AppCenterReactNative register];  // Initialize AppCenter

  [FIRApp configure];

  [RNFirebaseNotifications configure];

  [[FBSDKApplicationDelegate sharedInstance] application:application didFinishLaunchingWithOptions:launchOptions];

  #ifdef DEBUG
    [RNBranch useTestInstance];
  #endif // DEBUG
  [RNBranch initSessionWithLaunchOptions:launchOptions isReferrable:YES];

  // listener for Branch Deep Link data
  [[Branch getInstance] initSessionWithLaunchOptions:launchOptions andRegisterDeepLinkHandler:^(NSDictionary * _Nonnull params, NSError * _Nullable error) {
    // do stuff with deep link data (nav to page, display content, etc)
    NSLog(@"%@", params);
  }];

  NSURL *jsCodeLocation;

  #ifdef DEBUG
    jsCodeLocation = [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index" fallbackResource:nil];
  #else
    jsCodeLocation = [CodePush bundleURL];
  #endif // DEBUG

  RCTRootView *rootView = [[RCTRootView alloc] initWithBundleURL:jsCodeLocation moduleName:@"Fantastec" initialProperties:nil launchOptions:launchOptions];
  rootView.backgroundColor = [[UIColor alloc] initWithRed:1.0f green:1.0f blue:1.0f alpha:1];
  self.window = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
  UIViewController *rootViewController = [UIViewController new];
  rootViewController.view = rootView;
  self.window.rootViewController = rootViewController;
  [self.window makeKeyAndVisible];
  UIView* launchScreenView = [[[NSBundle mainBundle] loadNibNamed:@"LaunchScreen" owner:self options:nil] objectAtIndex:0];
  launchScreenView.frame = self.window.bounds;
  rootView.loadingView = launchScreenView;
  return YES;
}

- (void)application:(UIApplication *)application didReceiveLocalNotification:(UILocalNotification *)notification {
  [[RNFirebaseNotifications instance] didReceiveLocalNotification:notification];
}

- (void)application:(UIApplication *)application didReceiveRemoteNotification:(nonnull NSDictionary *)userInfo fetchCompletionHandler:(nonnull void (^)(UIBackgroundFetchResult))completionHandler {
  [[RNFirebaseNotifications instance] didReceiveRemoteNotification:userInfo fetchCompletionHandler:completionHandler];
  [[Branch getInstance] handlePushNotification:userInfo];
}

- (void)application:(UIApplication *)application didRegisterUserNotificationSettings:(UIUserNotificationSettings *)notificationSettings {
  [[RNFirebaseMessaging instance] didRegisterUserNotificationSettings:notificationSettings];
}

- (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDictionary<UIApplicationOpenURLOptionsKey,id> *)options {
  if (![RNBranch.branch application:app openURL:url options:options]) {
    // do other deep link routing for the Facebook SDK, Pinterest SDK, etc
  }
  if ([[FBSDKApplicationDelegate sharedInstance] application:app openURL:url options:options]) {
    return YES;
  }
  [[Branch getInstance] application:app openURL:url options:options];
  return NO; // fbsdk NO, branch.io YES
}

- (BOOL)application:(UIApplication *)application continueUserActivity:(NSUserActivity *)userActivity restorationHandler:(void (^)(NSArray *restorableObjects))restorationHandler {
  // handler for Universal Links
  return [RNBranch continueUserActivity:userActivity];
}

@end
```

But still getting https://github.com/facebook/react-native-fbsdk/issues/648#issuecomment-611253282 since going live with this change Thursday with 500+ installs.

I've checked https://github.com/BranchMetrics/react-native-branch-deep-linking-attribution/blob/master/examples/testbed_native_ios/ios/testbed_native_ios/AppDelegate.m out too but failed to gain clarity.

jdee:
@leotm We (I) need to clean up the docs on the openURL method. I noticed our Obj-C method snippet looked funny. What version of `react-native-branch` and for that matter `react-native` are you using? I'll rewrite a couple of methods for you.

leotm:
@jdee That would be much appreciated,
- "react-native": "0.61.5"
- "react-native-branch": "^4.4.0"

leotm:
Also I've been attempting to rectify the above with https://dashboard.branch.io/start/existing-users/ios?next=/start/existing-users/android OBJECTIVE-C instructions to the best of my ability taking into account how React-Native may vary

Screenshot 2020-04-22 at 21 36 22

Screenshot 2020-04-22 at 21 36 36

jdee:
In `application:didFinishLaunchingWithOptions:` don't call `[[Branch getInstance] initSession]`. Only call `[RNBranch initSessionWithLaunchOptions:isReferrable:]`.

Here's how I would write the openURL: method.
```obj-c
- (BOOL)application:(UIApplication *)app
            openURL:(NSURL *)url
            options:(NSDictionary<UIApplicationOpenURLOptionsKey,id> *)options
{
  // 1. Check if this is a Branch link
  // If using a custom domain (not *.app.link) be sure to set branch_universal_link_domains
  // in Info.plist or the return value of the following method will not be correct.
  // See https://help.branch.io/developers-hub/docs/ios-basic-integration#section-configure-infoplist
  // When updating to react-native-branch 5.x change RNBranch.branch to just RNBranch.
  if ([RNBranch.branch application:app openURL:url options:options]) {
    // If Branch handled it, we're done.
    return YES;
  }

  // 2. Check if this is a FB link
  if ([[FBSDKApplicationDelegate sharedInstance] application:app openURL:url options:options]) {
    // If Facebook handled it, we're done
    return YES;
  }

  // 3. Anything not handled by FB or Branch is passed to Linking and up to JS
  if ([RCTLinkingManager application:app openURL:url options:options]) {
    return YES;
  }

  // don't call [[Branch getInstance] openURL:] here

  // 4. If Linking somehow punted on the link, it's a good idea to pass it back to the OS (probably into a browser).
  return [[UIApplication sharedApplication] openURL:url];
}
```

jdee:
And there's no need to add `pod "Branch"` to your Podfile. `react-native-branch` is a wrapper around the native iOS Branch SDK. It pulls in that pod for you, with the right version we need. That all happens automatically with RN >= 0.60, when you add `use_native_modules!` to your Podfile. 

One point regarding branch.json that I think may not be clear in the docs. You have to add it to your Xcode project so that it is included in the app bundle for the SDK to find at runtime.

The best reference example app in this repo is probably webview_example, which is still using 4.4.0. That's a pure RN app (cross-platform).

If you continue to have issues, could you please contact support@branch.io? We'll help you get sorted out there.

leotm:
> And there's no need to add `pod "Branch"` to your Podfile. `react-native-branch` is a wrapper around the native iOS Branch SDK. It pulls in that pod for you, with the right version we need. That all happens automatically with RN >= 0.60, when you add `use_native_modules!` to your Podfile.

I thought so, I've been leaving it out from the Podfile noticing the auto-resolution in Podfile.lock 👍

I'll give this a proper shot tomorrow morning, sure thing thanks again for the help.

PR #561, opened 2020-03-27
Update native sdks

jdee:
This change updates to Android 5.0.0, iOS 0.32.0. This required a change to use the new `registerPlugin` method on Android. Calls to initSession in RNBranchModule.java will show up as deprecated for now. Those should be adjusted to use the new InitSessionBuilder.

This led to the discovery that the plugin versions were hard-coded in RNBranchConfig.m (iOS) and build.gradle (Android). That means that version 4.4.0 of this SDK reports itself to Branch as 4.3.0.

This change now passes the version as a preprocessor macro in the podspec (iOS) and loads the version from package.json in the build.gradle (Android).

Also note that RNBranchModule.getAutoInstance() was never a mandatory change. It was only needed to support keys from branch.json. I modified it in testbed_simple. If anyone is still using Branch.getAutoInstance() in their application class on Android, they will not report plugin version info to Branch.

There will be one more change for 3.x, since there we still support Carthage and have a RNBranch.xcodeproj for that. I'll probably have to update that in the Fastlane version_bump action that performs version bumps in this repo. I'm not sure if the project can load the package.json at build time.


ios/RNBranchConfig.m
```diff
@@ -20,7 +20,7 @@
 NSString * _Nonnull const RNBranchConfigDeferInitializationForJSLoadOption = @"deferInitializationForJSLoad";
 NSString * _Nonnull const RNBranchConfigEnableFacebookLinkCheck = @"enableFacebookLinkCheck";
 
-NSString * const RNBNC_PLUGIN_VERSION = @"4.3.0";
+NSString * const RNBNC_PLUGIN_VERSION = RNBRANCH_VERSION;
```
bklastaitis-branch: @jdee how can I test this? I tried checking out your branch, then form inside `testbed_simple`

- npm uninstall --save react-native-branch
- npm install --save /Users/benas.klastaitis/AndroidStudioProjects/plugins/react-native-branch-deep-linking-attribution/
- open testbed_simple in XCode and build but I get the following error.

/Users/benas.klastaitis/AndroidStudioProjects/plugins/react-native-branch-deep-linking-attribution/examples/testbed_simple/node_modules/react-native-branch/ios/RNBranchConfig.m:23:41: error: use of undeclared identifier 'RNBRANCH_VERSION'
NSString * const RNBNC_PLUGIN_VERSION = RNBRANCH_VERSION;

What am I missing?

jdee: @bklastaitis-branch You'll need to run `pod update`:

```bash
cd ios
bundle check || bundle install
bundle exec pod update
```

That will regenerate the Pods project to pass that definition through.

examples/testbed_simple/android/app/src/main/java/com/testbed_simple/MainApplication.java
```diff
@@ -40,6 +40,6 @@ public ReactNativeHost getReactNativeHost() {
   public void onCreate() {
     super.onCreate();
     SoLoader.init(this, /* native exopackage */ false);
-    Branch.getAutoInstance(this);
+    RNBranchModule.getAutoInstance(this);
```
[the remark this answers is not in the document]
jdee: Made that change. Only browser_example was missing it in the 4.x examples. Also updated the 3.x examples in #562 .

We no longer support Carthage with 4.x, so I also removed webview_example_carthage in this PR.

ios/RNBranchConfig.m
bklastaitis-branch: TY!

Issue #530, opened 2020-01-16
Have access to the raw URL clicked by user


zyskowsk:
We have a react native app with which we are having significant reliability issues when it comes to deep linking.  

We use branch links to pass invite tokens to the app, if the app cannot retreive the invite token the user is blocked and cannot advance. 

We have a significant number of users who follow a branch link when the app is already installed, the are never redirected. 

We have the following code called in the `componentDidMount` of our root component:

```js
  branchSubscribe = branch.subscribe(({ error, params }) => {
    // `+non_branch_link` handles incoming standard deeplinks and universal links.
    const url = params["+non_branch_link"] || params["~referring_link"];
    if (url) {
      redirectFromUrl(url);
    }

    if (error) {
      logWarning("Error from Branch: " + error);
      return;
    }
  });
};
```

The callback passed to `branch.subscribe` is never called when the device has zero network connectivity.  Even in cases where we have low connectivity, this callback is called unreliably. 

The raw link [is cached on the native side](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/blob/master/Branch-SDK/Branch-SDK/Branch.m#L764-L781), but it does not seem that this is exposed to RN through any channel that is not triggered by the response to the Branch Servers

`getLatestReferringParams` seems a bit more reliable, but the data it returns [still seems to be set after a request to the branch servers.](https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/blob/a9d8c47fbefbb23c45a67113887d6a1878fb340c/Branch-SDK/Branch-SDK/Networking/Requests/BranchOpenRequest.m#L188)

Is there any way to always have reliable access to the raw link that the user clicked ?

jamesholcomb:
I too am experiencing this using email verification tokens.  Additionally, even if the app has network connectivity, it may still timeout trying to connect to branch servers.

One solution would be for the `branch.subscribe()` callback to always return the native layer referring link on network errors.  That would allow the deep link account verification flow to continue.

echo-branch:
Adding a ticket for this one.

zyskowsk:
I think it could make sense to give a separate listener to pass the `referral_link` directly to the app when wanted. The case described above, we are perfectly happy to let branch function in the background, but we aren't relying on anything from branch for the app to function.

The approach we are taking in the short term is to user React Native Linking in parallel to branch:

```
[...]

PR #564, opened 2020-04-01
[SDK-779] BranchSubscriber

jdee:
Address #530 

This change introduces a BranchSubscriber class in JS. You can create a BranchSubscriber using `new BranchSubscriber` or use `branch.subscribe` as before. The following three code snippets do exactly the same thing:

```js
import { BranchSubscriber } from 'react-native-branch'

const subscriber = new BranchSubscriber({
  onOpenComplete: ({params, error, uri}) => { ... },
})
subscriber.subscribe()
```

```js
import branch from 'react-native-branch'

branch.subscribe({
  onOpenComplete: ({params, error, uri}) => { ... },
})
```

```js
import branch from 'react-native-branch'

branch.subscribe(({params, error, uri}) => { ... })
```

The `uri` parameter is the raw link that opened the app from the native layer. To be notified before the Branch SDK posts its open request, add an `onOpenStart` callback in the Object with `onOpenComplete`:

```js
const subscriber = new BranchSubscriber({
  onOpenStart: ({uri}) => { ... },
  onOpenComplete: ({params, error, uri}) => { ... },
})
subscriber.subscribe()
```

```js
branch.subscribe({
  onOpenStart: ({uri}) => { ... },
  onOpenComplete: ({params, error, uri}) => { ... },
})
```

No change is necessary for any app that doesn't require the onOpenStart callback.

The testbed_simple app has been modified to install from the local repo and adjusted to use the BranchSubscriber. The other example apps still point at NPM. See testbed_simple/src/Subscribe.js.

In basic testing on both Android and iOS, this works with URL opens. There are some cases where there is no URI/URL to report in the onOpenStart callback, particularly deferred deep links, but also perhaps things like Spotlight listings. For now, onOpenStart will be called, perhaps inconsistently, with a null `uri` argument in those cases. And there may be differences in the behavior on Android and iOS in these cases. For now, with these changes as they stand, the SDK will do its level best to notify the caller when it's about to open a URI, but no guarantees. This can be improved at need.

To run unit tests for JS and native iOS:

```bash
./runAllTests
```

Also squeezed in a bit of general repo cleanup. In particular, the scripts directory is gone. Old tests using ava have been moved to ava-tests. As a separate piece of work, I still want to convert the rest of those to jest and delete ava-tests.


android/src/main/java/io/branch/rnbranch/RNBranchModule.java
```diff
@@ -117,7 +122,8 @@ public static void reInitSession(Activity reactActivity) {
         Intent intent = reactActivity.getIntent();
         if (intent != null) {
             intent.putExtra("branch_force_new_session", true);
-            branch.reInitSession(reactActivity, referralInitListener);
+            notifyJSOfInitSessionStart(reactActivity, intent.getData());
```
bklastaitis-branch, line 125: Is the name `notifyJSOfInitSessionStart` misspelled?

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
```diff
@@ -232,7 +207,18 @@ private void generateLocalBroadcast(JSONObject referringParams,
             }
         }.init(reactActivity);
 
-        branch.initSession(referralInitListener, uri, reactActivity);
+        notifyJSOfInitSessionStart(reactActivity, uri);
+        branch.sessionBuilder(reactActivity).withCallback(referralInitListener).withData(uri).init();
```
bklastaitis-branch, line 211: I believe `sessionBuilder` is a static method, so shouldn't be called on an instance object

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
```diff
@@ -232,7 +207,18 @@ private void generateLocalBroadcast(JSONObject referringParams,
...
+    }
+
+    private static void notifyJSOfInitSessionStart(Context context, Uri uri) {
+        Intent broadcastIntent = new Intent(NATIVE_INIT_SESSION_STARTED_EVENT);
+        if (uri != null) {
+            broadcastIntent.putExtra(NATIVE_INIT_SESSION_STARTED_EVENT_URI, uri);
+        }
+
+        LocalBroadcastManager.getInstance(context).sendBroadcast(broadcastIntent);
+        Log.d(REACT_CLASS, "Sent session start broadcast for " + uri);
```
bklastaitis-branch, line 221: Should prolly replace `Log.d` with `PrefHelper.Debug` so it is only printed when logging is enabled.

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
```diff
@@ -309,8 +296,8 @@ public RNBranchModule(ReactApplicationContext reactContext) {
         return constants;
     }
 
-    private void forwardInitSessionFinishedEventToReactNative(ReactApplicationContext reactContext) {
-        mInitSessionEventReceiver = new BroadcastReceiver() {
+    private void forwardInitSessionEventsToReactNative(ReactApplicationContext reactContext) {
```
bklastaitis-branch, line 299: Since we are renaming the method anyway, could we rename it to `setupForwardingInitSessionEventsToReactNative`? (This name wouldn't imply an action and IMHO would reflect what the method does better.)

examples/testbed_simple/android/app/build.gradle
```diff
@@ -178,6 +178,9 @@ android {
 dependencies {
     implementation fileTree(dir: "libs", include: ["*.jar"])
     implementation "com.facebook.react:react-native:+"  // From node_modules
+    implementation 'com.google.firebase:firebase-appindexing:19.1.0'
+    implementation 'com.google.android.gms:play-services-ads-identifier:17.0.0'
```
bklastaitis-branch, line 182: I believe version 17 requires extra configuration steps , are you sure this works on Android? See warning `Google Play Services version 17+` for more information (https://help.branch.io/developers-hub/docs/android-basic-integration) If not, it's safe to return to v16+.

examples/testbed_simple/package.json
```diff
@@ -8,7 +8,7 @@
 	"dependencies": {
 		"react": "16.8.6",
 		"react-native": "0.60.3",
-		"react-native-branch": "^4.1.0"
+		"react-native-branch": "../.."
```
bklastaitis-branch, line 11: 👍

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
jdee: I don't think it is. This is a camel-case representation of: "notify JS of init session start." But it could be called something else.

examples/testbed_simple/android/app/build.gradle
jdee: I just added this since I was having trouble testing at first. I guess I misread 16+ as `>= 16` and assumed this was also correct and more current. I can remove this or correct it. It's probably more realistic to add these things to the examples.

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
jdee: Good suggestion. I agree.

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
jdee: :+1: Thanks for that suggestion.

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
jdee: :+1: Thanks. This seems to work in testing. I think C++ and Java are not as fussy as other languages about static/member calling. That is, a static member is always available to instances. Certainly true in C++. In Obj-C, e.g. this sort of error would fail to compile. I'll fix it.

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
bklastaitis-branch: haha I see! Yeah, maybe we can leave a comment above the method "notify JS of init session start."

I personally would name it `notifyJSInitSessionStarted` or `notifyJSThatInitSessionStarted` but naming is always subjective, and you work on the RN project most of the time anyway, so whatever you choose is fine by me :)

examples/testbed_simple/android/app/build.gradle
bklastaitis-branch: Interesting, running the test app worked for me too. Perhaps the extra configuration is needed only when importing `com.google.android.gms:play-services-ads` not `com.google.android.gms:play-services-ads-identifier` in which case I need to make a DOCS ticket.. But yeah, 16+ means the highest minor/patch version of version 16.

echo-branch:
The iOS side of this looks good!  Going to a do a bit of testing and then approve.

PR #589, opened 2020-07-08
[SDK-1024] customer event alias

jdee:
Added an `alias` property to BranchEvent, e.g.:

```js
const event = new BranchEvent(BranchEvent.VIEW_ITEM, buo, { alias: "Item Viewed" });
```

This is mapped to the `alias` property on iOS and `setCustomerEventAlias()` on Android.

This was branched from `SDK-1034/update-native-sdks` in order to use the latest SDKs. This can be merged after #587.


echo-branch:
Found an issue while testing.  BranchEvent on the JS has the alias, but it's not getting translated all the way to iOS.  Looks like there's a second layer of mapping done here.
https://github.com/BranchMetrics/react-native-branch-deep-linking-attribution/blob/master/src/BranchEvent.js#L163-L190

jdee:
@echo-branch Huh. I updated that and the unit tests for it. I guess those changes didn't get pushed? Thanks for catching that.

jdee:
Wait. Yes they did:

https://github.com/BranchMetrics/react-native-branch-deep-linking-attribution/pull/589/files#diff-de2b97386ae475e335b344e9841fefc8R127
https://github.com/BranchMetrics/react-native-branch-deep-linking-attribution/pull/589/files#diff-ce75a7b871231ae9fd6c7dea6bad3b14R193-R201

jdee:
Missed that one. Also added alias to standard event in testbed_simple for testing.

PR #593, opened 2020-07-20
[SDK-1050] Return an appropriate function for unsubscribe.

### LINKED ISSUE #590: branch unsubscribe isnt working
I'm using this configuration :

```
react-native: 0.61.5
react-native-branch: 4.3.0"
```
and using this code to subscribe to a branch callback, 

```
this._unsubscribeFromBranch = branch.subscribe(async ({ error, params }) => {
            this._unsubscribeFromBranch();
            this._unsubscribeFromBranch = null;
            this.branchCallback(error, params, sessionToken);
          });
```

and in the same component:

```
componentWillUnmount() {
    if (this._unsubscribeFromBranch) {
      this._unsubscribeFromBranch();
      this._unsubscribeFromBranch = null;
    }
  }
```
with this configuration the callback is been trigger, but when in android the back key is pressed and the app is sent to background (app closes, but isn't killed) after re enter the app, the callback is been trigger on app start, even if this line is executed:

`this._unsubscribeFromBranch();
`

is there a better way to unsubscribe all events after a branch.subscribe?

jdee:
Fix #590

Returning `subscriber.unsubscribe` from `branch.subscribe` somehow does not result in a working unsubscribe function. Changed to `return () => subscriber.unsubscribe()`, i.e. a function that calls that function. This consistently works in testing.

### DIFF
--- a/native-tests/ios/Podfile.lock
+++ b/native-tests/ios/Podfile.lock
@@ -184,7 +184,7 @@ PODS:
     - React-cxxreact (= 0.62.0)
     - React-jsi (= 0.62.0)
   - React-jsinspector (0.62.0)
-  - react-native-branch (5.0.0-beta.2):
+  - react-native-branch (5.0.0-rc.1):
     - Branch (= 0.34.0)
     - React
   - React-RCTActionSheet (0.62.0):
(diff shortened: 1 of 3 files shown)


jdee:
I think an alternative here might have been `return subscriber.unsubscribe.bind(subscriber)`. I think returning the bare `subscriber.unsubscribe` function lost the binding to the subscriber object. With this change, the subscriber is captured as part of the lambda returned.

PR #596, opened 2020-07-24
Rebuild testbed_simple

jdee:
Rebuilt testbed_simple from scratch using `react-native init`. Then just basically imported the existing source code for the app. This works fine now on iOS and Android and is the main testing app for this SDK. The changes are really just to the RN app scaffolding.


jdee:
Also made progress on webview_example, but there have been significant changes to react-navigation.

PR #598, opened 2020-07-25
Rebuild webview_example

jdee:
Similarly to #596, rebuilt webview_example from scratch using `react-native init`. Updating to the `@react-navigation` API from the legacy `react-navigation` package (see https://reactnavigation.org) meant significant source code changes in JS as well. This runs well now on both Android and iOS and provides a more realistic implementation scenario than testbed_simple.


echo-branch:
Also works on my machine!

PR #605, opened 2020-09-29
Move BranchConfig files and implementation to native SDKs

adrianyg7:
Move BranchConfig files and implementation to native SDKs


android/src/main/java/io/branch/rnbranch/RNBranchModule.java
```diff
@@ -733,12 +715,6 @@ private static Branch setupBranch(Context context) {
         if (!mInitialized) {
             Log.i(REACT_CLASS, "Initializing Branch SDK v. " + BuildConfig.VERSION_NAME);
 
-            RNBranchConfig config = new RNBranchConfig(context);
-
-            if (mUseDebug || config.getDebugMode()) branch.setDebug();
```
adrianyg7, line 738: `setDebug` is deprecated and currently no-op. Should we get rid of `mUseDebug` too?

ios/RNBranch.m
```diff
@@ -51,22 +50,14 @@ + (Branch *)branch
...
         static dispatch_once_t once = 0;
         dispatch_once(&once, ^{
-            RNBranchConfig *config = RNBranchConfig.instance;
-
-            // YES if either [RNBranch useTestInstance] was called or useTestInstance: true is present in branch.json.
-            BOOL usingTestInstance = useTestInstance || config.useTestInstance;
-            NSString *key = branchKey ?: config.branchKey ?: usingTestInstance ? config.testKey : config.liveKey;
-
-            if (key) {
+            if (branchKey) {
```
adrianyg7, line 53: Is there a scenario where `branchKey` has a non-null?
With the migration to native SDKs, branchKey will be set to whatever branch.json has or fallback to plist file.
If branchKey is set at this point, it will override json and plist. Is that ok?

ios/RNBranch.m
```diff
@@ -167,10 +138,10 @@ + (void)initSessionWithLaunchOptions:(NSDictionary *)launchOptions isReferrable:
     savedLaunchOptions = launchOptions;
     savedIsReferrable = isReferrable;
 
-    [self.branch registerPluginName:@"ReactNative" version:RNBNC_PLUGIN_VERSION];
+    [self.branch registerPluginName:@"ReactNative" version:RNBRANCH_VERSION];
 
     // Can't currently support this on Android.
-    // if (!deferInitializationForJSLoad && !RNBranchConfig.instance.deferInitializationForJSLoad) [self initializeBranchSDK];
+    // if (!deferInitializationForJSLoad && !BranchJsonConfig.instance.deferInitializationForJSLoad) [self initializeBranchSDK];
```
adrianyg7, line 144: Didn't remove these references so we can have a starting point in case this gets implemented in native SDKs

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
jdee: In this class? If it's not used because of this, then definitely.

ios/RNBranch.m
jdee: You're stumbling into some unreleased code. One thing that has been requested by RN users for a long time is to be able to set keys from JS, but that requires waiting for JS to load and deferring native initialization. Android has always posed limitations there. Unfortunately things like that have bitten us in the past when quietly released only for iOS. People assume they're supported on Android too, and then we have to make it so. So I've commented out some bits here and there, but it is basically possible to initialize the native iOS SDK from JS. That's what the `initializeBranch:` method is for around line 352.

jdee (commented):
Let's wait to merge this until we've released the native SDKs and can include the version increments here. Otherwise merging this breaks the staging branch.

PR #608, opened 2020-10-06
Add enableLogging to RNBranch

jkadamczyk:
RNBranch had missing `enableLogging` method, which is per docs, the advised way of debugging as `setDebug` got deprecated.

Currently using it through `patch-package` would be glad if it got merged.

Have a nice day!


jkadamczyk:
Tested it on device and works like a charm, also in combination with `useTestInstance`

PR #621, opened 2021-01-08
Add getLastAttributedTouchData on JS layer in React Native

justinfurstossbranch:
Add support for lastAttributedTouchDataWithAttributionWindow. 

This involved adding the method in `RNBranchModule.java` 
```
    @ReactMethod
    public void lastAttributedTouchDataWithAttributionWindow(int window, final Promise promise) {
        Branch branch = Branch.getInstance();
        branch.getLastAttributedTouchData(new ServerRequestGetLATD.BranchLastAttributedTouchDataListener() {
                @Override
                public void onDataFetched(JSONObject jsonObject, BranchError error) {
                    if (error == null) {
                        promise.resolve(jsonObject);
                    } else {
                        promise.reject(GENERIC_ERROR, error.getMessage());
                    }
                }
            }, window);
    }
```

As well as adding the method to support iOS builds in RNBranch.m
```
#pragma mark lastAttributedTouchDataWithAttributionWindow
RCT_EXPORT_METHOD(
                  lastAttributedTouchDataWithAttributionWindow:(NSNumber* __nonnull)window
                  resolver:(RCTPromiseResolveBlock)resolve
                  rejector:(__unused RCTPromiseRejectBlock)reject
                  ) {
    [self.class.branch lastAttributedTouchDataWithAttributionWindow:window.integerValue completion:^(BranchLastAttributedTouchData *r){
        resolve(r);
    }];
}

```
Tested with testbed_simple example project with adding a new button.

### DIFF
--- a/ios/RNBranch.m
+++ b/ios/RNBranch.m
@@ -402,6 +402,17 @@ RCT_EXPORT_METHOD(
     resolve([self.class.branch getFirstReferringParams]);
 }
 
+#pragma mark lastAttributedTouchDataWithAttributionWindow
+RCT_EXPORT_METHOD(
+                  lastAttributedTouchDataWithAttributionWindow:(NSNumber* __nonnull)window
+                  resolver:(RCTPromiseResolveBlock)resolve
+                  rejector:(__unused RCTPromiseRejectBlock)reject
+                  ) {
+    [self.class.branch lastAttributedTouchDataWithAttributionWindow:window.integerValue completion:^(BranchLastAttributedTouchData *r){
+        resolve(r);
+    }];
+}
+
 #pragma mark setIdentity
 RCT_EXPORT_METHOD(
                   setIdentity:(NSString *)identity
--- a/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
+++ b/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
@@ -438,7 +438,7 @@ public class RNBranchModule extends ReactContextBaseJavaModule {
         branch.getLastAttributedTouchData(new ServerRequestGetLATD.BranchLastAttributedTouchDataListener() {
                 @Override
                 public void onDataFetched(JSONObject jsonObject, BranchError error) {
-                    if (error != null) {
+                    if (error == null) {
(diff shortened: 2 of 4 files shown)


ios/RNBranch.m
```diff
@@ -402,6 +402,17 @@ - (BranchUniversalObject *)findUniversalObjectWithIdent:(NSString *)ident reject
...
 }
 
+#pragma mark lastAttributedTouchDataWithAttributionWindow
+RCT_EXPORT_METHOD(
+                  lastAttributedTouchDataWithAttributionWindow:(NSNumber* __nonnull)window
+                  resolver:(RCTPromiseResolveBlock)resolve
+                  rejector:(__unused RCTPromiseRejectBlock)reject
+                  ) {
+    [self.class.branch lastAttributedTouchDataWithAttributionWindow:window.integerValue completion:^(BranchLastAttributedTouchData *r){
+        resolve(r);
```
jdee, line 412: You have a promise rejection on Android but not here. Is it possible to get an error here?

jdee (approved):
:+1: LGTM. There can be small differences between the iOS and Android SDKs. If there's a way to detect any error on iOS (e.g., nil returned), I'd add the rejection, but otherwise looks fine.

ios/RNBranch.m
justinfurstossbranch: Correct me if I am wrong, but I think that would require a change to the native iOS SDK since https://github.com/BranchMetrics/ios-branch-deep-linking-attribution/blob/master/Branch-SDK/Branch.m#L1288 only returns the BranchLastAttributedTouchData object.

justinfurstossbranch: If you think the promise rejection is not an immediate/mandatory requirement for pushing this out, I'll be more than happy merge the PR. With that said, I would also be happy owning the native iOS change to support this and then make the RN change as well if so desired.

justinfurstossbranch: I know we have a client actively waiting on this release, so if you're comfortable, I will merge the PR. I can also own making the change on our native iOS SDK to support both LATD and Error object returns within the completion and update the RN SDK to handle said change for a future release.

Issue #669, opened 2021-08-20
Type definitions are wrong


mmclafferty:
Hello,
I am on the latest. react-native-branch 5.0.4 and ran into a few issues with typescript.

`createBranchUniversalObject` is supposed to return a promise.
The object it returns has a function called `showShareSheet` which is also supposed to return a promise with some data.

PR #697, opened 2022-02-05
Fix createBranchUniversalObject's TypeScript definition

v-fernandez:
createBranchUniversalObject is an async function, so it's TypeScript definition should be a function that returns `Promise<BranchUniversalObject>` instead of `BranchUniversalObject`.


v-fernandez:
Partly addresses #669

PR #712, opened 2022-05-20
Fix BranchUniversalObject showShareSheet return type definition

romanlitvin:
_showShareSheet_ method in BranchUniversalObject has incorrect return type. It indicates that function is sync and returns void. When function is async and returns Promise with a result.


gdeluna-branch:
Thanks! Will get this released soon.

PR #715, opened 2022-06-01
Change ~creation_source type to be a number

### LINKED ISSUE #709: ~creation_source type should be number
I noticed the typing for `~creation_source` is wrong. It should be number instead of string. This can be verified easily by creating [a deeplink from API.](https://help.branch.io/developers-hub/docs/deep-linking-api#creating-a-deep-linking-url). Putting this into `branch.subscribe` and console logging `event.params` will tell you that `~creation_source` is a number. You can also do a check like this and confirm that check returns false.

```
event.params[~creation_source] === '0' // false
```

Also by the document, these are numbers.

> Where the link was created (0 API , 1 Quick Link, 2 SDK, 3 iOS SDK , 4 Android SDK , 5 Web SDK, 6 Dynamic, 7 Third party)

Let me know if you want me to submit a PR to fix this.

vincent-paing:
Fixes #709

I've tested locally by changing the type manually within `node_modules` folder but you might want to run an e2e or integration test to confirm it doesn't break anything.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -17,7 +17,7 @@ export interface BranchParams {
   "~tags"?: string[];
   "~campaign"?: string;
   "~stage"?: string;
-  "~creation_source"?: string;
+  "~creation_source"?: number;
   "~referring_link"?: string;
   "~id"?: string;
   "+match_guaranteed": boolean;


jf-branch:
Hey @vincent-paing ~creation_source is an internally used parameter, and it is actually set on the native side, i.e Android = 'android', iOS = 'iOS' we'll actually want to remove it from the index.d.ts.

echo-branch (approved):

PR #732, opened 2022-08-09
[SDK-1562] Add callback to setIdentity

nsingh-branch:
### Description
Adds a new method, `setIdentityWithPromise`, which sets the identity/developer_identity and has a promise with returns the first referring params. There already is a `setIdentity` functions which works the same and will still exist. However, it is void/does not return, which is why adding a new method which does.

The native SDKs have two versions of setIdentity, one with a callback and one thats void. This new method uses the one with a callback.

### Usage/Examples
**Existing `setIdentity()`**
```
branch.setIdentity("user123")
```

**New `setIdentityWithPromise()`**
```
let result = branch.setIdentityWithPromise("user456");
    result.then((params) => {
      console.log('setIdentityWithPromise Result: ', params)
});
```

### DIFF
--- a/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
+++ b/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
@@ -446,13 +446,13 @@ public class RNBranchModule extends ReactContextBaseJavaModule {
     }
 
    @ReactMethod
-    public void setIdentityWithPromise(String identity, Promise promise) {
+    public void setIdentityAsync(String identity, Promise promise) {
         Branch branch = Branch.getInstance();
         branch.setIdentity(identity, new BranchReferralInitListener() {
             @Override
             public void onInitFinished(JSONObject referringParams, BranchError error) {
                 if (error != null) {
-                    promise.reject("RNBranch::Error::setIdentityWithPromise failed", error.getMessage());
+                    promise.reject("RNBranch::Error::setIdentityAsync failed", error.getMessage());
                 } else {
                     promise.resolve(convertJsonToMap(referringParams));
                 }
--- a/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
+++ b/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
@@ -443,16 +443,16 @@ public class RNBranchModule extends ReactContextBaseJavaModule {
     public void setIdentity(String identity, Promise promise) {
         Branch branch = Branch.getInstance();
 
-          branch.setIdentity(identity, new BranchReferralInitListener() {
-                    @Override
-                    public void onInitFinished(JSONObject referringParams, BranchError error) {
-                        if (error != null) {
-                            promise.reject("RNBranch::Error::setIdentity failed", error.getMessage());
-                        } else {
-                            promise.resolve(convertJsonToMap(branch.getFirstReferringParams()));
-                        }
-                    }
-                });
+        branch.setIdentity(identity, new BranchReferralInitListener() {
+            @Override
+            public void onInitFinished(JSONObject referringParams, BranchError error) {
+                if (error != null) {
+                    promise.reject("RNBranch::Error::setIdentity failed", error.getMessage());
+                } else {
+                    promise.resolve(convertJsonToMap(convertJsonToMap(referringParams)));
+                }
+            }
+        });
     }
 
     @ReactMethod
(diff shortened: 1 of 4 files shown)


ios/RNBranch.m
```diff
@@ -403,8 +403,17 @@ - (BranchUniversalObject *)findUniversalObjectWithIdent:(NSString *)ident reject
 #pragma mark setIdentity
 RCT_EXPORT_METHOD(
                   setIdentity:(NSString *)identity
+                  resolver:(RCTPromiseResolveBlock)resolve
+                  rejecter:(__unused RCTPromiseRejectBlock)reject
                   ) {
-    [self.class.branch setIdentity:identity];
+    [self.class.branch setIdentity: identity withCallback:^(NSDictionary *params, NSError *error) {
+        if (!error) {
+            resolve([self.class.branch getFirstReferringParams]);
```
jf-branch, line 411: @nsingh-branch Don't we have to do the same to RNBranchModule.java https://github.com/BranchMetrics/react-native-branch-deep-linking-attribution/blob/master/android/src/main/java/io/branch/rnbranch/RNBranchModule.java#L442 and have the promise.resolve or promise.reject?

gdeluna-branch:
Can you add more to the description? What did the calls look like previously and what have they changed to now?

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
```diff
@@ -440,9 +440,10 @@ public void onDataFetched(JSONObject jsonObject, BranchError error) {
     }
 
     @ReactMethod
-    public void setIdentity(String identity) {
+    public void setIdentity(String identity, Promise promise) {
         Branch branch = Branch.getInstance();
         branch.setIdentity(identity);
+        promise.resolve(convertJsonToMap(branch.getFirstReferringParams()));
```
gdeluna-branch, line 446: do we need to reject here as well?

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
```diff
@@ -440,9 +440,10 @@ public void onDataFetched(JSONObject jsonObject, BranchError error) {
     }
 
     @ReactMethod
-    public void setIdentity(String identity) {
+    public void setIdentity(String identity, Promise promise) {
         Branch branch = Branch.getInstance();
         branch.setIdentity(identity);
```
gdeluna-branch, line 445: Do we need to pass in anything for the second argument `BranchReferralInitListener` ?

src/index.d.ts
```diff
@@ -309,7 +309,7 @@ interface Branch {
   getLatestReferringParams: (synchronous?: boolean) => Promise<BranchParams>;
   getFirstReferringParams: () => Promise<BranchParams>;
   lastAttributedTouchData: (attributionWindow?: number) => Promise<BranchParams>;
-  setIdentity: (identity: string) => void;
+  setIdentity: (identity: string) => Promise<BranchParams>;
```
gdeluna-branch, line 312: Since this changes the type, code that exists today may get a warning since it is no longer void and the return type isn't being used. Can you confirm this doesn't break anything?

nsingh-branch: Would it make sense to keep the void function and add a new function `setIdentityWithPromise` that returns the promise/callback?

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
```diff
@@ -445,6 +445,21 @@ public void setIdentity(String identity) {
...
     }
 
+   @ReactMethod
+    public void setIdentityAsync(String identity, Promise promise) {
```
JagadeeshKaricherla-branch, line 449: you dont need to pass a promise as param.
`Promise.resolve` , `Promise.reject` should work.

nsingh-branch: Didn't we test this and find it was necessary for it to work?

JagadeeshKaricherla-branch: can you check if java allows `Promise.resolve`  ( promise with a uppercase P ) if not I think we are good.

JagadeeshKaricherla-branch (approved):

android/src/main/java/io/branch/rnbranch/RNBranchModule.java
JagadeeshKaricherla-branch: I think e started testing but couldn’t fully run that day. Yaa if it is necessary, lets keep it.

Issue #725, opened 2022-07-11
customData in track event crash if it's not string


vincent-paing:
## Definition

The type for `customData` is set to `AnyDataType` (Ref:https://github.com/BranchMetrics/react-native-branch-deep-linking-attribution/blob/master/src/index.d.ts#L74),

```ts
type AnyDataType =
  | string
  | boolean
  | number
  | null
  | undefined
  | AnyDataArray
  | AnyDataObject;
 ```
 
 However using anything other than string will cause the app to crash in Android. You can observe the following error in native log
 
 ```
Exception in native call
com.facebook.react.bridge.UnexpectedNativeTypeException: Value for fb_success cannot be cast from Double to String
    at com.facebook.react.bridge.ReadableNativeMap.checkInstance(ReadableNativeMap.java:140)
    at com.facebook.react.bridge.ReadableNativeMap.getNullableValue(ReadableNativeMap.java:128)
    at com.facebook.react.bridge.ReadableNativeMap.getString(ReadableNativeMap.java:162)
    at io.branch.rnbranch.RNBranchModule.createBranchEvent(RNBranchModule.java:710)
    at io.branch.rnbranch.RNBranchModule.logEvent(RNBranchModule.java:479)
    at java.lang.reflect.Method.invoke(Native Method)
    at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:372)
    at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:188)
    at com.facebook.react.bridge.queue.NativeRunnable.run(Native Method)
    at android.os.Handler.handleCallback(Handler.java:938)
    at android.os.Handler.dispatchMessage(Handler.java:99)
    at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:27)
    at android.os.Looper.loop(Looper.java:223)
    at com.facebook.react.bridge.queue.MessageQueueThreadImpl$4.run(MessageQueueThreadImpl.java:226)
    at java.lang.Thread.run(Thread.java:923)
 ```

## Reproducible code

```ts
new BranchEvent(BranchEvent.Search, undefined, {
    customData: {
      fb_success: 0,
    },
  }).logEvent();
```

## Cause

[...]

PR #740, opened 2022-09-21
[SDK-1517] [ReactNative] customData in track event crash if it's not string

gdeluna-branch:
An issue was raised https://github.com/BranchMetrics/react-native-branch-deep-linking-attribution/issues/725 wherein calling a `BranchEvent` with `customParams` with non-string values would throw an uncaught exception. 

After review, we decided to define the typing as  
 `customData?: Record<string, string>;`

This typing will avoid the mismatch in the native layer.

To test, install the beta containing this change.
https://www.npmjs.com/package/react-native-branch/v/5.6.0-beta.0
via
`npm install — save react-native-branch@beta`

### DIFF
--- a/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
+++ b/android/src/main/java/io/branch/rnbranch/RNBranchModule.java
@@ -773,7 +773,14 @@ public class RNBranchModule extends ReactContextBaseJavaModule {
             ReadableMapKeySetIterator it = customData.keySetIterator();
             while (it.hasNextKey()) {
                 String key = it.nextKey();
-                event.addCustomDataProperty(key, customData.getString(key));
+                
+                ReadableType readableType = customData.getType(key);
+                if(readableType == ReadableType.String) {
+                    event.addCustomDataProperty(key, customData.getString(key));
+                }
+                else{
+                    Log.w(REACT_CLASS, "customData values must be strings. Value for property " + key + " is not a string type.");
+                }
             }
         }
 
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -71,7 +71,7 @@ export class BranchEvent {
   description?: string;
   searchQuery?: string;
   alias?: string;
-  customData?: Record<string, AnyDataType>;
+  customData?: Record<string, string>;
 
   /**
    * Standard Add to Cart event


JagadeeshKaricherla-branch:
@gdeluna-branch : I think we should also fix at the place where we are trying to convert.
https://github.com/BranchMetrics/react-native-branch-deep-linking-attribution/blob/b490d5e0ee0bc815d9abd76299ff0f84a16d5c91/android/src/main/java/io/branch/rnbranch/RNBranchModule.java#L776

Basically check the type and warn if the type is not string

`import com.facebook.react.bridge.ReadableType;`


```
while (it.hasNextKey()) {
    String key = it.nextKey();
    ReadableType type = customData.getType(key);
    if (type == ReadableType.String) {
        event.addCustomDataProperty(key, customData.getString(key));
    } else {
        Log.w(REACT_CLASS, "[Branch] customData values must be strings. Value for property " + key + " is not a string type.");
    }
}
```

We already do a similar check in the BranchEvent here https://github.com/BranchMetrics/react-native-branch-deep-linking-attribution/blob/f69ed47e1742edc8b7082982db934ff14ed42425/src/BranchEvent.js#L189

gdeluna-branch:
> @gdeluna-branch : I think we should also fix at the place where we are trying to convert.
> 
> https://github.com/BranchMetrics/react-native-branch-deep-linking-attribution/blob/b490d5e0ee0bc815d9abd76299ff0f84a16d5c91/android/src/main/java/io/branch/rnbranch/RNBranchModule.java#L776
> 
> Basically check the type and warn if the type is not string
> 
> `import com.facebook.react.bridge.ReadableType;`
> 
> ```
> while (it.hasNextKey()) {
>     String key = it.nextKey();
>     ReadableType type = customData.getType(key);
>     if (type == ReadableType.String) {
>         event.addCustomDataProperty(key, customData.getString(key));
>     } else {
>         Log.w(REACT_CLASS, "[Branch] customData values must be strings. Value for property " + key + " is not a string type.");
>     }
> }
> ```
> 
> We already do a similar check in the BranchEvent here
> 
> https://github.com/BranchMetrics/react-native-branch-deep-linking-attribution/blob/f69ed47e1742edc8b7082982db934ff14ed42425/src/BranchEvent.js#L189

Done. This is the JS side type warning
image


This is the warning on the native side, 
```
 WARN  [Branch] customMetadata values must be strings. Value for property fb_success has type number.
```
Change published in beta https://www.npmjs.com/package/react-native-branch/v/5.6.0-beta.1

JagadeeshKaricherla-branch:
@gdeluna-branch : missing below import statement
`import com.facebook.react.bridge.ReadableType;`

JagadeeshKaricherla-branch (approved):

PR #753, opened 2022-11-01
Fix BranchUniversalObject#generateShortUrl parameter type definition

zeallat:
I modified it to be closer to the requirements of the actual js code. All arguments have default values, so it seemed right to give them as optionals.

https://github.com/BranchMetrics/react-native-branch-deep-linking-attribution/blob/dfa570b073c4fddba189a1604382f0093f514059/src/branchUniversalObject.js#L74


JagadeeshKaricherla-branch (approved):
LGTM

gdeluna-branch (approved):

PR #799, opened 2023-03-07
[SDK-1593] Implementation for JS deferred init

gdeluna-branch:
## Reference
SDK-1539 -- New API for plugin ready state and callback

## Summary
To address a race condition where apps don't receive Branch params on cold starts, an opt in fix will defer loading the native iOS/Android layer until signaled by this plugin in `subscribe()`.

This can be enabled by creating a `branch.json` file with the contents:
  ```
  {
    "deferInitForPluginRuntime": true
  }
  ```
Android: Place this file in your src/main/assets folder
iOS: Add this file through Xcode, File -> Add Files to "YourProject.xcodeproj"
and add to Copy Bundle Resources for each target that inits the Branch SDK.

## Motivation

This is one of the more common issues raised in our issues. Before, the workaround would be to set a TTL for cached params, but that is not consistent enough. This implementation will wait until the subscriber is ready.

## Type Of Change

- [✅] Bug fix (non-breaking change which fixes an issue)
- [✅] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [✅] This change requires a documentation update

## Testing Instructions
Follow the above instructions and see that on cold starts, your app still receives branch params. Implemented in the testbed app for demonstration.


echo-branch (approved):
Confirmed proper behavior on iOS with and without flag enabled.

PR #957, opened 2024-02-26
[SDK-2249] Expose the setDMAParamsForEEA method

nsingh-branch:
## Reference
SDK-2249 -- Expose the setDMAParamsForEEA method

## Summary
Expose the setDMAParamsForEEA method in the JS layer and bump the native SDK versions.

## Motivation
So clients can utilize the method in React Native and stay compliant.

## Type Of Change

- [x] New feature (non-breaking change which adds functionality)

## Testing Instructions

Call the method like `Branch.setDMAParamsForEEA(true, true, true);` and observe the following request payloads for the respective DMA fields.


ios/RNBranch.m
echo-branch, line 1: setDebug was removed from iOS.

echo-branch:
Was able to set the DMA parameters on iOS and see them in logs. So it looks like the happy path works.

echo-branch (approved):
Changes look good in my testing. I think the automation failure is maybe due to the removal of the deprecated API?

echo-branch (approved):

PR #967, opened 2024-03-27
[SDK-2326] Add setAPIUrl to the JS layer

nsingh-branch:
## Reference
SDK-2326 --  Expose the setAPIUrl method in the JS layer

## Summary
Added the setAPIUrl to the JS layer now that it exists on both of the native layers.

## Motivation

Bring the React Native SDK to parity with out iOS and Android native SDKs.

## Type Of Change

- [x] New feature (non-breaking change which adds functionality)

## Testing Instructions

Call the new method with a cusrtom API URL. To have the initial open/install use the endpoint, it needs to be set early. 
Calling `branch.setAPIUrl('https://api3.branch.io')` in the Testbed's `componentDidMount` works in that case.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -164,6 +164,17 @@ class Branch {
       console.warn("setDMAParamsForEEA: Unable to set DMA params.");
     }
   };
+
+  /*** Set the Branch API's base URL ***/
+  setAPIURL = (apiUrl) => {
+    if (!apiUrl || typeof apiUrl !== "string" || !apiUrl.startsWith("http")) {
+      console.warn(
+        "setAPIURL: Invalid URL. URL must be a non-empty string starting with 'http'."
+      );
+      return;
+    }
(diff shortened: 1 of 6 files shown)


src/index.js
```diff
@@ -164,6 +164,17 @@ class Branch {
       console.warn("setDMAParamsForEEA: Unable to set DMA params.");
     }
   };
+
+  /*** Set the Branch API's base URL (eg. "https://api2.branch.io") ***/
+  setAPIUrl = (apiUrl) => {
```
gdeluna-branch, line 169: I recommend removing the JavaScript API and moving this to the native JSON api. The JS <> native runtime switch will be a race condition against this setting which must be called before session init. 

Instead, the json can be read on the native init's start up.

nsingh-branch:
Closing in favor of using branch.json

PR #1036, opened 2024-10-03
fix(types): allow custom fields in BranchLinkControlParams

zakwalters:
## Reference
SDK-XXXX -- <TITLE>.

## Summary

Updates the `BranchLinkControlParams` to allow custom entries.

## Motivation

The documentation says this is how custom data, such as `referring_user_id` should be added to deep links, but the types didn't allow for this. Ignoring the types and testing link generation showed that values were passed through successfully, so no change was needed on the code side.

From my testing, all values passed here (strings, numbers, booleans, nulls) were stringified, so I went with `Record<string, string>` as the type so that users will expect strings on the other side. I also found that passing an explicit `undefined` would drop the key entirely from the result, so I thought the `Record` approach was better than `[custom: string]: string | undefined`, where users might expect the key to be present but with no value, or a `null` when they handle the deep link.

## Type Of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Testing Instructions

---

As an aside, your CONTRIBUTING.md still points to an issue tracker, even though issues seem to have been disabled for this repo.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -268,14 +268,14 @@ export interface BranchLinkProperties {
   tags?: string[];
 }
 
-export interface BranchLinkControlParams {
+export type BranchLinkControlParams = {
   $fallback_url?: string;
   $desktop_url?: string;
   $ios_url?: string;
   $ipad_url?: string;
   $android_url?: string;
   $samsung_url?: string;
-}
+} & Record<string, string>;
 
 interface BranchShareSuccess {
   completed: true;


zakwalters:
@nsingh-branch not sure if the PRs are monitored given the lack of response, tagging for some visibility

PR #1038, opened 2024-10-08
Expose native layer Integration Validator

rob-gioia-branch:
## Reference
No ticket. Done to make it easier for clients to test & validate their Branch integration.

## Summary
Exposed the native Android and iOS validators so that customers can make a single call at the JS layer to run the validator on whatever platform their app is running on. 

## Motivation
Currently, there is no way to run the Integration Validator from JS code. We have to tell clients to edit the native code, which isn't ideal and gets pushback from some customers. Others are confused as to how to do that. With this change, clients can run one line of code branch.validateSDKIntegration() to use the Integration Validator. 

## Type Of Change
- [ ] New feature (non-breaking change which adds functionality)

## Testing Instructions

- Run this line of code in your sample / test bed app: branch.validateSDKIntegration();

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -48,5 +48,4 @@ dependencies {
     implementation 'androidx.localbroadcastmanager:localbroadcastmanager:1.0.0'
     implementation 'com.facebook.react:react-native:+' // From node_modules
     api 'io.branch.sdk.android:library:5.12.4'
-    implementation 'com.google.android.gms:play-services-base:18.5.0'
 }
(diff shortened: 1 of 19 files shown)


android/build.gradle
```diff
@@ -48,4 +48,5 @@ dependencies {
     implementation 'androidx.localbroadcastmanager:localbroadcastmanager:1.0.0'
     implementation 'com.facebook.react:react-native:+' // From node_modules
     api 'io.branch.sdk.android:library:5.12.4'
+    implementation 'com.google.android.gms:play-services-base:18.5.0'
```
gdeluna-branch, line 51: Does this need to be distributed?

rob-gioia-branch: I just removed it and the IV code still worked, so I committed the removal as we don't seem to need it. 

Thanks Gabe, good catch!

gdeluna-branch (approved):

gdeluna-branch:
There are some merge conflicts

gdeluna-branch (approved):

